PR #169, opened 2015-03-23
Support starting before logind.

Kurt-von-Laven:
Previously we required the event recorder daemon to start after logind
so we could conveniently inhibit shutdown on startup. The
instrumentation daemon starts before logind to more conveniently record
logins. The instrumentation daemon should start after the event recorder
daemon so that it shuts down before the instrumentation daemon. This is
necessary so that the instrumentation daemon can record the shutdown
event in time for the event recorder daemon to persist it to disk.
However, this results in a systemd ordering cycle unless we break one of
the ordering dependencies. This change breaks this would-be cycle by
removing the requiremment that the event recorder daemon start after
logind. Instead of inhibiting shutdown on startup, the event recorder
daemon now inhibits shutdown on startup or once the login manager starts
up, whichever happens last.

[endlessm/eos-sdk#2906]


ptomato:
Since the `AX_PYTHON_MODULE` macro is copied from an outside source, consider looking for another already existing macro that does what you want, or submitting the changes back to the original.

Kurt-von-Laven:
I didn't find another macro that enforces versioning of Python modules (only of Python itself). I can submit a pull request against https://github.com/peti/autoconf-archive/blob/master/m4/ax_python_module.m4 with whatever version of this passes our internal CR.

daemon/emer-main.c
```diff
@@ -187,18 +192,23 @@ on_bus_acquired (GDBusConnection *system_bus,
     {
       g_error ("Could not export metrics interface on system bus: %s",
                error->message);
+      g_error_free (error);
```
ptomato: Not necessary, because `g_error` aborts the program.

m4/ax-python-module.m4
```diff
@@ -4,13 +4,14 @@
 #
 # SYNOPSIS
 #
-#   AX_PYTHON_MODULE(modname[, fatal])
+#   AX_PYTHON_MODULE(module_name, min_version)
```
ptomato: Upon further reading I would suggest checking the version in a separate macro. It can be a very simple macro if you use this macro to determine that the module is present already. Also consider doing the comparison from within Python in order to gain access to a dedicated function for comparing versions.

Something like this:
```m4
AC_DEFUN([AX_PYTHON_MODULE_VERSION], [
    AX_PYTHON_MODULE([$1])
    AC_MSG_CHECKING([for version $2 of $1])
    echo "import sys, $1; from distutils.version import StrictVersion; sys.exit(StrictVersion($1.__version__) >= StrictVersion('$2'))" > conftest.py
    $PYTHON conftest.py >/dev/null 2>&1
    AS_IF([test $? -eq 0], [], [
        AC_MSG_RESULT([no])
        AC_MSG_ERROR([You need at least version $2 of the $1 Python module.])
    ])
    AC_MSG_RESULT([yes])
])
```

daemon/emer-boot-id-provider.c
```diff
@@ -105,18 +104,16 @@ emer_boot_id_provider_get_property (GObject    *object,
       break;
 
     default:
-      G_OBJECT_WARN_INVALID_PROPERTY_ID (object, property_id, pspec);
+      G_OBJECT_WARN_INVALID_PROPERTY_ID (self, property_id, pspec);
```
ptomato: This should now have a cast: `G_OBJECT (self)`.

My personal preference would be to have this function's signature take a `GObject` since `get_property` is an override of a `GObject` vfunc, but that's up to you.

Kurt-von-Laven: Just looked upstream and apparently GLib itself always does this the way it was written originally. Does dropping [this commit](https://github.com/endlessm/eos-event-recorder-daemon/commit/e52f8798b630f7c9d170c9094972a80bde9cddab) entirely seem like a reasonable way of addressing this comment?

ptomato: Sounds good.

daemon/emer-main.c
Kurt-von-Laven: Doh!

cosimoc:
@Kurt-von-Laven Revise ready

PR #172, opened 2015-03-30
Update timestamps on shutdown.

Kurt-von-Laven:
When the PrepareForShutdown signal is sent with parameter TRUE,
indicating that we are about to shutdown, update the relative and
absolute timestamps in the persistent cache's boot offset metadata
file. This makes the metrics' systems time-keeping algorithm more robust
in the event that the timestamp update that is triggered when the
persistent cache is finalized doesn't make it to disk before the system
shuts down.

[endlessm/eos-sdk#1558]

### DIFF
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -930,6 +930,21 @@ finally:
 }
 
 static void
+update_timestamps (EmerDaemon *self)
+{
+  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
+
+  GError *error = NULL;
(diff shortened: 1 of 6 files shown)


Kurt-von-Laven:
The first two commits: "Permit NULL offset out parameter" and "Update timestamps on shutdown" are intended to be cherry-picked to eos2.3. The third commit ("Remove an unnecessary disk read") is lower priority but could also be worthwhile to cherry-pick since it presumably could improve startup performance a teeny bit.

Kurt-von-Laven:
@cosimoc CR ready!

Kurt-von-Laven:
@cosimoc Fixed the indentation issues. CR ready!

PR #174, opened 2015-03-31
Send metrics on first try.

Kurt-von-Laven:
Currently we don't send any metrics on the first try on a new image
because we notice that there is no network send file, write a default
send file and bail. Instead of bailing out, we now plow ahead and send
metrics regardless of whether we were able to find a network send file.

[endlessm/eos-sdk#3003]

### DIFF
--- a/daemon/emer-network-send-provider.c
+++ b/daemon/emer-network-send-provider.c
@@ -191,16 +191,19 @@ reset_network_send_data (EmerNetworkSendProvider *self)
                   error->message);
       g_error_free (error);
     }
+
+  priv->send_number = 0;
+  priv->data_cached = TRUE;
 }
 
-static gboolean
+static void
 read_network_send_data (EmerNetworkSendProvider *self)
 {
   EmerNetworkSendProviderPrivate *priv =
     emer_network_send_provider_get_instance_private (self);
 
   if (priv->data_cached)
-    return TRUE;
+    return;
 
   GError *error = NULL;
   if (!g_key_file_load_from_file (priv->key_file, priv->path, G_KEY_FILE_NONE,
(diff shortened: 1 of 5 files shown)


daemon/emer-network-send-provider.c
```diff
@@ -191,15 +191,19 @@ reset_network_send_data (EmerNetworkSendProvider *self)
...
 }
 
-static gboolean
+static void
 read_network_send_data (EmerNetworkSendProvider *self)
 {
   EmerNetworkSendProviderPrivate *priv =
     emer_network_send_provider_get_instance_private (self);
+
   if (priv->data_cached)
```
plward11: @Kurt-von-Laven Does a `gboolean` get initialized to `FALSE` if you just declare `gboolean my_gboolean`?

ptomato: Nope, it gets initialized to garbage, which is likely to be unequal to 0 and therefore `TRUE`

Kurt-von-Laven: In most circumstances there is no guarantee that data of any type is initialized in C. gcc automatically initializes static variables to their default values (e.g., NULL, 0).

I have a vague memory of @ptomato mentioning that `g_object_new` initializes all variables in the private struct to their default values unless they are otherwise specified, but [its documentation](https://developer.gnome.org/gobject/stable/gobject-The-Base-Object-Type.html#g-object-new) only claims this about construction parameters. I wonder if the only reason that it works in practice is that the kernel initializes new pages to 0 when we fire up a new process.

plward11: In that case, it should be initialized.

ptomato: GObject definitely initializes the private struct to zeros intentionally. It may not be documented, though. Feel free to open a GNOME bug saying it should be documented ;-)

Kurt-von-Laven: Just ran make check 100 times in a for loop, and indeed `priv->data_cached` is always initialized to `0`. For the record though, uninitialized memory is highly likely to be 0 (largely because the kernel initializes new pages to 0 to prevent one process from reading another's memory). 0 is so much more likely than other values that a clever academic demonstrated that performance could be improved by guessing that any given attempt to read RAM would return 0 and backpedaling on the off chance that it was something else.

plward11: What about initializing `data_cached` to `FALSE` in `emer_network_send_provider_init` for style purposes?

ptomato: Here is the proof that it is intentionally filled with zeros: https://git.gnome.org/browse/glib/tree/gobject/gtype.c#n1843

Kurt-von-Laven: Thanks; had trouble finding that. Personally, I tend to find initializing values explicitly easier to read as evidenced by https://github.com/endlessm/eos-event-recorder-daemon/blob/master/daemon/emer-daemon.c#L1409-L1410 and other places in the metrics system. We work with a lot of different programming languages, and I have a hard time keeping track of which ones automatically initialize variables under which circumstances. It's also more obviously wrong if a variable is explicitly initialized to 0 but should have been initialized to some other value. I don't have very strong sentiments about it now that we are very certain that it does the right thing either way.

ptomato: I used to initialize members explicitly but that tended to lead to twenty identical lines:
```c
priv->some_widget = NULL;
priv->thing_widget = NULL;
priv->whatever_widget = NULL;
...
```

Kurt-von-Laven: Filed https://bugzilla.gnome.org/show_bug.cgi?id=747147.

Kurt-von-Laven: I propose that we initialize members explicitly if and only if upstream refuses to mention that `g_object_new` zeroes out members of the private struct in its official documentation, which would suggest that they consider it an implementation detail that may be changed on a whim in future releases. Regardless, it's a pre-existing issue if it's an issue at all. @plward11 do you have any other comments on this pull request, or can we merge it?

Kurt-von-Laven: Cosimo points out that upstream already officially mentions that `g_object-new` zeroes out members of the private struct at https://developer.gnome.org/gobject/stable/chapter-gobject.html#gobject-instantiation.

plward11: Nah, I was concerned about style, but Cosimo cleared that up.

PR #176, opened 2015-04-08
Document use of subdir-objects.

Kurt-von-Laven:
[endlessm/eos-sdk#3029]

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -38,6 +38,8 @@ AC_CONFIG_SRCDIR([daemon/emer-daemon.h])
 # no-portability suppresses warnings about syntax specific to GNU make
 # parallel-tests specifies that we use the new parallel-running test harness.
 # Unlike serial-tests, this option is accepted by Automake 1.11.
+# subdir-objects ensures forward compatibility with Automake 2.0 and can be
+# removed when this behavior becomes the default.
 # tar-ustar is required because there may be files whose entire paths exceed
 # 99 characters.
 AM_INIT_AUTOMAKE([-Wall -Wno-portability foreign 1.11 parallel-tests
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -35,9 +35,10 @@ AC_INIT([EOS Event Recorder Daemon], [0.0.0],
 # Verify that the source directory can be found
 AC_CONFIG_SRCDIR([daemon/emer-daemon.h])
 # Initialize Automake: enable all warnings and do not insist on GNU standards.
-# no-portability suppresses warnings about syntax specific to GNU make
+# no-portability suppresses warnings about syntax specific to GNU make.
 # parallel-tests specifies that we use the new parallel-running test harness.
 # Unlike serial-tests, this option is accepted by Automake 1.11.
+# subdir-objects mirrors the source tree in the build directory.
```
ptomato: That's not quite what subdir-objects does. In any case I'd prefer a rationale for using it rather than a repeat of the documentation. e.g. "subdir-objects ensures forward compatibility with Automake 2.0 and can be removed when this behaviour becomes the default."

cosimoc: @Kurt-von-Laven I agree with this suggestion.
The PR looks fine otherwise.

Kurt-von-Laven: Will fix, but I am a little confused now. I thought the reason specifying subdir-objects ensures forward compatibility is that the behavior is no longer going to be the default in Automake 2.0.

ptomato: In Automake 2.0 subdir-objects will always be turned on, with no option to turn it off, as far as I know. The warning telling you to enable it is so that you fix any breakage now instead of having it happen by surprise when you upgrade to 2.0.

Kurt-von-Laven: Ah hah! Thanks for the explanation.

PR #177, opened 2015-05-02
Upgrade from Python 2 to Python 3.

Kurt-von-Laven:
[endlessm/eos-sdk#3038]


m4/ax_python_module.m4
```diff
@@ -4,13 +4,15 @@
 #
 # SYNOPSIS
 #
-#   AX_PYTHON_MODULE(modname[, fatal])
+#   AX_PYTHON_MODULE(modname[, fatal, python])
```
ptomato: It would be better to leave this macro as-is, and instead set `$PYTHON` in `configure.ac`, since the macro already provides you that option to override the python executable.

Something like:
```m4
AC_PATH_PROG([PYTHON], [python3 python])
AX_PROG_PYTHON_VERSION([3.0], [], [AC_MSG_ERROR([You must have Python 3.])])
```
Checking for a Python 3 executable named `python` and using [`ax_prog_python_version.m4`](https://github.com/djmitche/autoconf-archive/blob/master/m4/ax_prog_python_version.m4) to make sure it is not Python 2, I'm not sure is necessary though. You might be able to get away with just `AC_PATH_PROG([PYTHON], [python3])`.

configure.ac
```diff
@@ -72,8 +72,9 @@ EMER_REQUIRED_MODULES="$EOS_METRICS_REQUIREMENT \
                        $SOUP_REQUIREMENT \
                        $UUID_REQUIREMENT"
 AC_SUBST(EMER_REQUIRED_MODULES)
+
 # This indicates a required dbus mocking interface for unit testing.
-AX_PYTHON_MODULE_VERSION([dbusmock], 0.10)
+AX_PYTHON_MODULE_VERSION([dbusmock], 0.10, python3)
```
ptomato: This needs to move down in order to pick up the results of `AC_PATH_PROG` for Python. It should probably go in the "Required libraries" section, so it was in the wrong place anyway.

tests/Makefile.am.inc
```diff
@@ -124,7 +124,7 @@ DBUSDAEMON_LOG_COMPILER = $(top_srcdir)/tests/launch-mock-dbus-tests.sh
 AM_DBUSDAEMON_LOG_FLAGS = -k --verbose
 LOG_COMPILER = gtester
 AM_LOG_FLAGS = -k --verbose
-PY_LOG_COMPILER = python
+PY_LOG_COMPILER = python3
```
ptomato: `$(PYTHON)`

tests/daemon/test-daemon.c
```diff
@@ -107,7 +107,7 @@ static void
 start_mock_logind_service (Fixture *fixture)
 {
   fixture->logind_mock = g_subprocess_new (G_SUBPROCESS_FLAGS_STDOUT_PIPE, NULL,
-                                           "python", "-m", "dbusmock",
+                                           "python3", "-m", "dbusmock",
```
ptomato: Not sure what's the best way to pick up the detected Python executable here. Perhaps compile with `-DPYTHON_PATH="\"$(PYTHON)\""` and use `PYTHON_PATH` here?

m4/ax_python_module.m4
Kurt-von-Laven: Just to confirm: I should change this only if my upstream PR is rejected?

ptomato: Correct.

configure.ac
```diff
@@ -153,6 +151,9 @@ AS_IF([test -f $EVENT_RECORDER_SERVER_XML],
 # ------------------
 PKG_CHECK_MODULES([EOS_EVENT_RECORDER_DAEMON], [$EMER_REQUIRED_MODULES])
 
+# This indicates a required dbus mocking interface for unit testing.
+AX_PYTHON_MODULE_VERSION([dbusmock], 0.10)
```
ptomato: Should this not include a `[python3]` argument?

ptomato: I guess not, since `python3` is the default.

ptomato:
**CR pass**

PR #179, opened 2015-05-13
Support specifying max cache size via file.

Kurt-von-Laven:
Read the maximum persistent cache size out of the key file
/etc/metrics/cache-size.conf rather than hard-coding it. Introduce a
cache size provider to abstract away the details of reading from the key
file. This functionality will allow us to increase the maximum size of
the persistent cache for the clients we send on the imminent bus trip
without affecting any other clients.

[endlessm/eos-sdk#3151]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Build products
 /eos-event-recorder-daemon-0.0.0
 /tests/daemon/test-boot-id-provider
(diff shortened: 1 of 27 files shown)


Kurt-von-Laven:
Only "Rename permissionsdir to configdir" and "Support specifying max cache size via file" are intended to be cherry-picked for the bus trip.

PR #180, opened 2015-05-13
Allow persistent cache to read own config file.

Kurt-von-Laven:
[endlessm/eos-sdk#3151]


Kurt-von-Laven:
This entire pull request is intended to be cherry-picked for the bus trip.

PR #181, opened 2015-05-14
Flush to persistent cache when uploading disabled.

Kurt-von-Laven:
If we can't upload to the network, we should flush to disk instead to
avoid overflowing our in-memory buffers.

[endlessm/eos-sdk#3151]


Kurt-von-Laven:
This entire pull request is intended to be cherry-picked.

Kurt-von-Laven:
@cosimoc, FYI.

PR #182, opened 2015-05-18
Fix eos-select-metrics-env script.

Kurt-von-Laven:
Add permissions_file back as a shared variable, and correct sed syntax.

[endlessm/eos-sdk#3124]

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -51,6 +51,7 @@ noinst_PROGRAMS =
 
 persistentcachedir=$(localstatedir)/cache/metrics/
 configdir=$(sysconfdir)/metrics/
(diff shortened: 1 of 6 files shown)


Kurt-von-Laven:
@rshuler FYI, here is a fix for the issue you spotted. Thanks again for noticing this.

cosimoc:
Pretty sure this should refer https://github.com/endlessm/eos-sdk/issues/3151 instead of https://github.com/endlessm/eos-sdk/issues/3124, but looks good.
@Kurt-von-Laven please adjust the status of those stories accordingly :-)

Kurt-von-Laven:
Oops! I guess there's nothing we can do about the commit messages referring to the wrong issue now, but I fixed the labels.

PR #184, opened 2015-05-27
Add emer_daemon_upload_events.

Kurt-von-Laven:
The new method allows consumers of the API to trigger a network upload
of all events currently stored on the client. Also, add a convenience
script that calls this new method. Finally, add a local server for
testing purposes and greatly improve test coverage of the daemon.

[endlessm/eos-sdk#3168]


Kurt-von-Laven:
All the commits up to and including "Add emer_daemon_upload_events." are intended to be cherry-picked (and only these commits). Note that GitHub sometimes displays commits in the incorrect order, so please use `git log` instead if you are interested in seeing which commits I'm referring to.

PR #188, opened 2015-06-15
Overhaul persistent cache API.

Kurt-von-Laven:
Eliminate SingularEvent, AggregateEvent, and SequenceEvent types. Use
variants instead for all three to eliminate a lot of code
near-duplication. Introduce a persistent ring buffer (circular file) to
allow the persistent cache to be efficiently partially removed from.
Increase persistent cache size to from 90 KiB to 10 MB to take advantage
of the new circular file. Also leverage the circular file to only delete
events once they've been delivered. Set out errors rather than logging
them internally within the persistent cache. This allows more helpful
error messages to be reported when users run eos-upload-metrics and
minimizes the number of log statements.

[endlessm/eos-sdk#599]

### DIFF
--- a/daemon/emer-circular-file.c
+++ b/daemon/emer-circular-file.c
@@ -1,0 +1,820 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*- */
+
+/* Copyright 2015 Endless Mobile, Inc. */
+
+/*
+ * This file is part of eos-event-recorder-daemon.
+ *
+ * eos-event-recorder-daemon is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * eos-event-recorder-daemon is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with eos-event-recorder-daemon.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+#include "emer-circular-file.h"
+
+#include <gio/gio.h>
+#include <glib.h>
+
+#include "shared/metrics-util.h"
+
+#define METADATA_GROUP_NAME "metadata"
+#define MAX_SIZE_KEY "max_size"
+#define SIZE_KEY "size"
+#define HEAD_KEY "head"
+
+typedef struct _EmerCircularFilePrivate
+{
+  GFile *data_file;
+  GKeyFile *metadata_key_file;
+
+  GByteArray *write_buffer;
+
+  guint64 max_size;
+  guint64 size;
+  goffset head;
+} EmerCircularFilePrivate;
+
+static void emer_circular_file_initable_iface_init (GInitableIface *iface);
+
+G_DEFINE_TYPE_WITH_CODE (EmerCircularFile, emer_circular_file, G_TYPE_OBJECT,
+                         G_ADD_PRIVATE (EmerCircularFile)
+                         G_IMPLEMENT_INTERFACE (G_TYPE_INITABLE, emer_circular_file_initable_iface_init))
+
+enum
+{
+  PROP_0,
+  PROP_PATH,
+  PROP_MAX_SIZE,
+  NPROPS
+};
+
+static GParamSpec *emer_circular_file_props[NPROPS] = { NULL, };
+
+static gchar *
+get_metadata_filepath (EmerCircularFile *self)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  gchar *data_filepath = g_file_get_path (priv->data_file);
+  gchar *metadata_filepath =
+    g_strconcat (data_filepath, METADATA_EXTENSION, NULL);
+  g_free (data_filepath);
+
+  return metadata_filepath;
+}
+
+static gboolean
+save_metadata_file (EmerCircularFile *self,
+                    GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  gchar *metadata_filepath = get_metadata_filepath (self);
+  gboolean save_succeeded =
+    g_key_file_save_to_file (priv->metadata_key_file, metadata_filepath, error);
+  g_free (metadata_filepath);
+
+  return save_succeeded;
+}
+
+static gboolean
+add_to_size (EmerCircularFile *self,
+             guint64           delta,
+             GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  guint64 new_size = priv->size + delta;
+  g_key_file_set_uint64 (priv->metadata_key_file, METADATA_GROUP_NAME, SIZE_KEY,
+                         new_size);
+  if (!save_metadata_file (self, error))
+    return FALSE;
+
+  priv->size = new_size;
+  return TRUE;
+}
+
+static gboolean
+set_metadata (EmerCircularFile *self,
+              guint64           size,
+              goffset           head,
+              GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  g_key_file_set_uint64 (priv->metadata_key_file, METADATA_GROUP_NAME, SIZE_KEY,
+                         size);
+  g_key_file_set_int64 (priv->metadata_key_file, METADATA_GROUP_NAME, HEAD_KEY,
+                        head);
+
+  if (!save_metadata_file (self, error))
+    return FALSE;
+
+  priv->size = size;
+  priv->head = head;
+  return TRUE;
+}
+
+static gboolean
+read_disk_bytes (EmerCircularFile *self,
+                 guint8           *buffer,
+                 gsize             num_bytes,
+                 gsize             max_size,
+                 GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  if (num_bytes == 0)
+    return TRUE;
+
+  GFileInputStream *file_input_stream =
+    g_file_read (priv->data_file, NULL /* GCancellable */, error);
+  if (file_input_stream == NULL)
+    return FALSE;
+
+  GInputStream *input_stream = G_INPUT_STREAM (file_input_stream);
+  gsize head_to_end = max_size - priv->head;
+  gsize bytes_head = MIN(num_bytes, head_to_end);
+  gsize bytes_start = num_bytes - bytes_head;
+  if (bytes_start > 0)
+    {
+      gpointer copy_to = buffer + bytes_head;
+      gboolean read_succeeded =
+        g_input_stream_read_all (input_stream, copy_to, bytes_start,
+                                 NULL /* bytes read */, NULL /* GCancellable */,
+                                 error);
+      if (!read_succeeded)
+        goto handle_failed_read;
+    }
+
+  GSeekable *seekable = G_SEEKABLE (file_input_stream);
+  gboolean seek_succeeded =
+    g_seekable_seek (seekable, priv->head, G_SEEK_SET, NULL /* GCancellable */,
+                     error);
+  if (!seek_succeeded)
+    goto handle_failed_read;
+
+  gboolean read_succeeded =
+    g_input_stream_read_all (input_stream, buffer, bytes_head,
+                             NULL /* bytes read */, NULL /* GCancellable */,
+                             error);
+  g_object_unref (file_input_stream);
+  return read_succeeded;
+
+handle_failed_read:
+  g_object_unref (file_input_stream);
+  return FALSE;
+}
+
+/* Returns the size of a length-encoded buffer excluding any truncated trailing
+ * element. Assumes each element in the buffer is preceded by its length in
+ * bytes encoded as a little-endian guint64. Also assumes that each length,
+ * element pair is concatenated to the next.
+ */
+static guint64
+get_trimmed_size (const guint8 *buffer,
+                  gsize         num_bytes)
+{
+  guint64 curr_pos = 0;
+  while ((curr_pos + sizeof (curr_pos)) < num_bytes)
+    {
+      guint64 little_endian_elem_size = *(const guint64 *) (buffer + curr_pos);
+      guint64 elem_size =
+        swap_bytes_64_if_big_endian (little_endian_elem_size);
+      guint64 next_pos = curr_pos + sizeof (curr_pos) + elem_size;
+      if (next_pos > num_bytes)
+        break;
+      curr_pos = next_pos;
+    }
+
+  return curr_pos;
+}
+
+/* Replace the data file with the contents of the given buffer, and update the
+ * metadata file accordingly. In the event of an error, the circular file may
+ * instead be left logically empty, but the data and metadata files will never
+ * be left in an inconsistent state.
+ */
+static gboolean
+overwrite (EmerCircularFile *self,
+           gconstpointer     buffer,
+           gsize             num_bytes,
+           guint64           prev_max_size,
+           GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  /* The size of the data file may not exceed the maximum size stored in the
+   * metadata file.
+   */
+  if (priv->max_size > prev_max_size)
+    g_key_file_set_uint64 (priv->metadata_key_file, METADATA_GROUP_NAME,
+                           MAX_SIZE_KEY, priv->max_size);
+  if (!set_metadata (self, 0, 0, error))
+    return FALSE;
+
+  gboolean write_succeeded =
+    g_file_replace_contents (priv->data_file, buffer, num_bytes, NULL, FALSE,
+                             G_FILE_CREATE_NONE, NULL, NULL, error);
+  if (!write_succeeded)
+    return FALSE;
+
+  /* We only need to update the maximum size if we didn't already. */
+  if (priv->max_size < prev_max_size)
+    g_key_file_set_uint64 (priv->metadata_key_file, METADATA_GROUP_NAME,
+                           MAX_SIZE_KEY, priv->max_size);
+
+  return add_to_size (self, num_bytes, error);
+}
+
+/* Change the maximum size of the circular file from prev_max_size to
+ * priv->max_size. If the new maximum is less than the amount of data currently
+ * in the buffer, then any data that doesn't fit will be removed. Also
+ * reorganizes the circular file so that its head is at the start of the file.
+ */
+static gboolean
+resize (EmerCircularFile *self,
+        guint64           prev_max_size,
+        GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  if (prev_max_size == priv->max_size)
+    return TRUE;
+
+  gsize bytes_to_read = MIN (priv->size, priv->max_size);
+  guint8 buffer[bytes_to_read];
+  gboolean read_succeeded =
+    read_disk_bytes (self, buffer, bytes_to_read, prev_max_size, error);
+  if (!read_succeeded)
+    return FALSE;
+
+  guint64 new_size = get_trimmed_size (buffer, bytes_to_read);
+  return overwrite (self, buffer, new_size, prev_max_size, error);
+}
+
+static gboolean
+continue_reading_from_start (EmerCircularFile *self,
+                             GInputStream     *input_stream,
+                             guint8           *partially_filled_buffer,
+                             gsize             total_bytes_to_read,
+                             gsize             bytes_read_end,
+                             GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  if (bytes_read_end == total_bytes_to_read)
+    return TRUE;
+
+  GSeekable *seekable = G_SEEKABLE (input_stream);
+  goffset curr_position = g_seekable_tell (seekable);
+  if (curr_position != priv->max_size)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA,
+                   "Circular file has a physical size of %" G_GOFFSET_FORMAT
+                   "bytes, but expected physical size to be %" G_GUINT64_FORMAT
+                   " bytes.", curr_position, priv->max_size);
+      return FALSE;
+    }
+
+  gboolean seek_succeeded =
+    g_seekable_seek (seekable, 0, G_SEEK_SET, NULL /* GCancellable */, error);
+  if (!seek_succeeded)
+    return FALSE;
+
+  guint8 *copy_to = partially_filled_buffer + bytes_read_end;
+  gsize bytes_remaining = total_bytes_to_read - bytes_read_end;
+  gsize bytes_read_start;
+  gboolean read_succeeded =
+    g_input_stream_read_all (input_stream, copy_to, bytes_remaining,
+                             &bytes_read_start, NULL /* GCancellable */, error);
+  if (!read_succeeded)
+    return FALSE;
+
+  if (bytes_remaining != bytes_read_start)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA,
+                   "Circular file is shorter than expected. Reached end of "
+                   "file at byte %" G_GSIZE_FORMAT ".", bytes_read_start);
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
+read_elem_size (EmerCircularFile *self,
+                GInputStream     *input_stream,
+                guint64          *elem_size,
+                GError          **error)
+{
+  guint64 little_endian_elem_size;
+  gsize bytes_read;
+  gboolean read_succeeded =
+    g_input_stream_read_all (input_stream,
+                             &little_endian_elem_size,
+                             sizeof (little_endian_elem_size),
+                             &bytes_read,
+                             NULL /* GCancellable */,
+                             error);
+  if (!read_succeeded)
+    return FALSE;
+
+  read_succeeded =
+    continue_reading_from_start (self,
+                                 input_stream,
+                                 (guint8 *) &little_endian_elem_size,
+                                 sizeof (little_endian_elem_size),
+                                 bytes_read,
+                                 error);
+  if (!read_succeeded)
+    return FALSE;
+
+  *elem_size = swap_bytes_64_if_big_endian (little_endian_elem_size);
+  return TRUE;
+}
+
+static gboolean
+append_elem_to_array (EmerCircularFile *self,
+                      GInputStream     *input_stream,
+                      GPtrArray        *elem_array,
+                      guint64           elem_size,
+                      GError          **error)
+{
+  guint8 elem_data[elem_size];
+  gsize bytes_read;
+  gboolean read_succeeded =
+    g_input_stream_read_all (input_stream, elem_data, elem_size, &bytes_read,
+                             NULL /* GCancellable */, error);
+  if (!read_succeeded)
+    return FALSE;
+
+  read_succeeded =
+    continue_reading_from_start (self, input_stream, elem_data, elem_size,
+                                 bytes_read, error);
+  if (!read_succeeded)
+    return FALSE;
+
+  GBytes *elem = g_bytes_new (elem_data, elem_size);
+  g_ptr_array_add (elem_array, elem);
+  return TRUE;
+}
+
+static void
+set_path (EmerCircularFile *self,
+          const gchar      *data_filepath)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  priv->data_file = g_file_new_for_path (data_filepath);
+}
+
+static void
+set_max_size (EmerCircularFile *self,
+              guint64           max_size)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  priv->max_size = max_size;
+}
+
+static void
+emer_circular_file_set_property (GObject      *object,
+                                 guint         property_id,
+                                 const GValue *value,
+                                 GParamSpec   *pspec)
+{
+  EmerCircularFile *self = EMER_CIRCULAR_FILE (object);
+
+  switch (property_id)
+    {
+    case PROP_PATH:
+      set_path (self, g_value_get_string (value));
+      break;
+
+    case PROP_MAX_SIZE:
+      set_max_size (self, g_value_get_uint64 (value));
+      break;
+
+    default:
+      G_OBJECT_WARN_INVALID_PROPERTY_ID (object, property_id, pspec);
+    }
+}
+
+static void
+emer_circular_file_finalize (GObject *object)
+{
+  EmerCircularFile *self = EMER_CIRCULAR_FILE (object);
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  g_clear_object (&priv->data_file);
+  g_clear_pointer (&priv->metadata_key_file, g_key_file_unref);
+  g_clear_pointer (&priv->write_buffer, g_byte_array_unref);
+
+  G_OBJECT_CLASS (emer_circular_file_parent_class)->finalize (object);
+}
+
+static void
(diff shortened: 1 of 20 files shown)


Kurt-von-Laven:
@cosimoc I'm not moving endlessm/eos-sdk#599 into CR ready yet because I don't think we should merge the code I am putting on display here just yet. I am mainly interested in your thoughts on the high level ideas presented here though, in particular, the introduction of the circular file and the refactor of the persistent cache API. All the tests pass at this point, but I will be cleaning up and breaking the code into smaller commits in the meantime.

daemon/emer-circular-file.h
```diff
@@ -0,0 +1,97 @@
...
+
+#define EMER_IS_CIRCULAR_FILE_CLASS(klass) \
+  (G_TYPE_CHECK_CLASS_TYPE ((klass), \
+  EMER_TYPE_CIRCULAR_FILE))
+
+#define EMER_CIRCULAR_FILE_GET_CLASS(obj) \
+  (G_TYPE_INSTANCE_GET_CLASS ((obj), \
+  EMER_TYPE_CIRCULAR_FILE, EmerCircularFileClass))
+
+#define METADATA_EXTENSION ".metadata"
```
cosimoc: Should probably not be in a public header since it's not namespaced.

cosimoc: Actually nevermind, we're not in the library here

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+ */
+gboolean
+emer_circular_file_remove (EmerCircularFile *self,
+                           guint64           token,
+                           GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  g_return_val_if_fail (token <= priv->size, FALSE);
```
cosimoc: Either always consistently check in all public methods for this or never. I don't think you should, since this is not a library.

ptomato:
Did you select the name "circular file" as a [pun](https://en.wiktionary.org/wiki/circular_file)? Either way, I found it humorous.

daemon/emer-circular-file.c
cosimoc: Also in this case it would return `FALSE` without setting the error

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+      if (next_data_bytes > data_bytes_to_read)
+        break;
+
+      gboolean append_succeeded =
+        append_elem_to_array (self, input_stream, elem_array, elem_size, error);
+      if (!append_succeeded)
+        goto handle_failed_read;
+
+      curr_data_bytes = next_data_bytes;
+      curr_disk_bytes += sizeof (elem_size) + elem_size;
```
cosimoc: Should add a comment here explaining that the `sizeof (elem_size)` is to include the length, which is only present in the data on disk, like commented on IRC.

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+  if (file_output_stream == NULL)
+    return FALSE;
+
+  GOutputStream *output_stream = G_OUTPUT_STREAM (file_output_stream);
+  GSeekable *seekable = G_SEEKABLE (file_output_stream);
+
+  goffset tail = (priv->head + priv->size) % priv->max_size;
+  guint64 space_available_at_tail = priv->max_size - tail;
+  gsize bytes_tail;
+  if (priv->write_buffer->len <= space_available_at_tail)
```
cosimoc: Since you use it later already, and this logic is not straightforward, I would save e.g. `gsize excess_data = priv->write_buffer->len - space_available_at_tail` and then check `if (excess_data <= 0)` here.

Kurt-von-Laven:
@ptomato I didn't know that meaning of circular file, but that's pretty great.

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+emer_circular_file_append (EmerCircularFile *self,
+                           gconstpointer     elem,
+                           guint64           elem_size)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  guint64 elem_size_on_disk = sizeof (elem_size) + elem_size;
+  guint64 total_size = priv->size + priv->write_buffer->len + elem_size_on_disk;
+  if (total_size > priv->max_size)
```
cosimoc: I feel it's a little inconsistent that this is the only API that does not have a GError when failing, but it makes sense looking at the code. I feel at least the `remove` and `purge` APIs should be consistent with this one.

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+
+handle_failed_read:
+  g_propagate_error (error, local_error);
+  return FALSE;
+}
+
+static void
+emer_circular_file_initable_init (GInitableIface *iface)
+{
+  iface->init = emer_circular_file_may_fail_init;
```
cosimoc: Nitpick: rename this to `emer_circular_file_initable_init` since it's implicit that it might fail.

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+    g_key_file_load_from_file (priv->metadata_key_file, metadata_filepath,
+                               G_KEY_FILE_NONE, &local_error);
+  g_free (metadata_filepath);
+  if (!load_succeeded)
+    {
+      if (!g_error_matches (local_error, G_KEY_FILE_ERROR,
+                            G_KEY_FILE_ERROR_NOT_FOUND) &&
+          !g_error_matches (local_error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+        {
+          g_propagate_error (error, local_error);
```
cosimoc: Could use `goto handle_failed_read;`

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+  guint64 little_endian_size =
+    g_key_file_get_uint64 (priv->metadata_key_file, METADATA_GROUP_NAME,
+                           SIZE_KEY, &local_error);
+  if (local_error != NULL)
+    goto handle_failed_read;
+
+  priv->size = swap_bytes_64_if_big_endian (little_endian_size);
+  if (priv->size > priv->max_size)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA, "Logical size "
```
cosimoc: Could also set the local error here and go to the common cleanup function for consistency.

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+  goffset little_endian_key =
+    g_key_file_get_int64 (priv->metadata_key_file, METADATA_GROUP_NAME,
+                          HEAD_KEY, &local_error);
+  if (local_error != NULL)
+    goto handle_failed_read;
+
+  priv->head = swap_bytes_64_if_big_endian (little_endian_key);
+  if (priv->head < 0 || priv->head >= priv->max_size)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA, "Pointer to "
```
cosimoc: See comment above

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+                             GInputStream     *input_stream,
+                             guint8           *partially_filled_buffer,
+                             gsize             total_bytes_to_read,
+                             gsize             bytes_read_end,
+                             GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  if (bytes_read_end < total_bytes_to_read)
```
cosimoc: Could you return early at the beginning of this function when `bytes_read_end >= total_bytes_to_read`?

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+    emer_circular_file_get_instance_private (self);
+
+  guint64 little_endian_max_size = swap_bytes_64_if_big_endian (priv->max_size);
+  /* The size of the data file may not exceed the maximum size stored in the
+   * metadata file.
+   */
+  if (priv->max_size > prev_max_size)
+    g_key_file_set_uint64 (priv->metadata_key_file, METADATA_GROUP_NAME,
+                           MAX_SIZE_KEY, little_endian_max_size);
+  if (!set_metadata (self, 0, 0, error))
```
cosimoc: `set_metadata()` will already set the `MAX_SIZE_KEY` in the keyfile to the value you pass (0 in this case). Isn't the above call not useful?

cosimoc: Ooh nevermind. I was confused between `MAX_SIZE_KEY` and `SIZE_KEY`

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+    return FALSE;
+
+  gchar *data_filepath = g_file_get_path (priv->data_file);
+  gboolean write_succeeded =
+    g_file_set_contents (data_filepath, buffer, num_bytes, error);
```
cosimoc: You could consider using `g_file_replace_contents()` with the `GFile` directly instead.

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  guint64 little_endian_max_size = swap_bytes_64_if_big_endian (priv->max_size);
+  /* The size of the data file may not exceed the maximum size stored in the
+   * metadata file.
+   */
+  if (priv->max_size > prev_max_size)
+    g_key_file_set_uint64 (priv->metadata_key_file, METADATA_GROUP_NAME,
+                           MAX_SIZE_KEY, little_endian_max_size);
```
cosimoc: On the other hand I'm not at all sure you need to do the `swap_bytes_64_if_big_endian()` dance before saving these values to the keyfile.

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+                 guint8           *buffer,
+                 gsize             num_bytes,
+                 gsize             max_size,
+                 GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  if (num_bytes == 0)
+    return TRUE;
```
cosimoc: Won't this case already be taken care in the normal flow of the code below?

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+  gsize bytes_head;
+  if (num_bytes <= head_to_end)
+    {
+      bytes_head = num_bytes;
+    }
+  else
+    {
+      bytes_head = head_to_end;
+      gpointer copy_to = buffer + bytes_head;
+      gsize bytes_start = num_bytes - bytes_head;
```
cosimoc: See other comment I posted about this calculation getting clearer if saved before the `if` block.

daemon/emer-circular-file.c
```diff
@@ -0,0 +1,825 @@
...
+             GError          **error)
+{
+  EmerCircularFilePrivate *priv =
+    emer_circular_file_get_instance_private (self);
+
+  guint64 new_size = priv->size + delta;
+  guint64 little_endian_size = swap_bytes_64_if_big_endian (new_size);
+  g_key_file_set_uint64 (priv->metadata_key_file, METADATA_GROUP_NAME, SIZE_KEY,
+                         little_endian_size);
+  gboolean save_succeeded = save_metadata_file (self, error);
```
cosimoc: For consistency with the function below, you could just make this `if (!save_metadata_file(self, error)) return FALSE;`

daemon/emer-daemon.c
```diff
@@ -1432,31 +1361,13 @@ emer_daemon_class_init (EmerDaemonClass *klass)
...
-                      "The number of events allowed to be stored in the "
-                      "sequence metric buffer",
-                      -1, G_MAXINT, DEFAULT_METRICS_PER_BUFFER_MAX,
-                      G_PARAM_CONSTRUCT_ONLY | G_PARAM_WRITABLE |
-                      G_PARAM_STATIC_STRINGS);
+  emer_daemon_props[PROP_MAX_BYTES_BUFFERED] =
+    g_param_spec_ulong ("max-bytes-buffered", "Max bytes buffered",
+                        "The maximum number of bytes of event data that may be "
+                        "buffered in memory. Does not include overhead.",
+                        0ul, G_MAXULONG, 100000ul /* 100 kB */,
```
cosimoc: Could use the constant defined at the top for this (or is it a different value?)

daemon/emer-daemon.c
```diff
@@ -417,131 +497,95 @@ handle_http_response (SoupSession         *http_session,
...
-  for (gint i = 0; list_of_aggregates[i] != NULL; i++)
-    g_variant_builder_add_value (aggregates_builder, list_of_aggregates[i]);
-  g_free (list_of_aggregates);
+  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
 
-  for (gint i = 0; list_of_sequences[i] != NULL; i++)
-    g_variant_builder_add_value (sequences_builder, list_of_sequences[i]);
-  g_free (list_of_sequences);
+  gsize curr_num_variants = 0;
+  for (gsize curr_bytes = 0; curr_num_variants < priv->variant_array->len;
```
cosimoc: I find it a little bit confusing to use a different variable in the loop than the one declared here...

daemon/emer-circular-file.c
Kurt-von-Laven: Normally yes, but a cruel and vindictive programmer might call `emer_circular_file_new ("foo", 10, NULL)`, causing a metadata file to be written but not a data file, followed by `g_object_unref (circular_file)` and `emer_circular_file_new ("foo", 20, error)` to trigger a resize.

daemon/emer-daemon.c
```diff
@@ -417,131 +497,95 @@ handle_http_response (SoupSession         *http_session,
...
-      g_variant_builder_init (&event_values_builder, G_VARIANT_TYPE ("a(xmv)"));
-      for (gint j = 0; j < curr_sequence->num_event_values; ++j)
-        {
-          EventValue *curr_event_value = curr_sequence->event_values + j;
-          g_variant_builder_add (&event_values_builder, "(xmv)",
-                                 curr_event_value->relative_timestamp,
-                                 curr_event_value->auxiliary_payload);
-        }
+  gsize curr_read_bytes = 0;
+  for (gsize i = 0; i< num_variants; i++)
```
cosimoc: Missing space before `<`

daemon/emer-circular-file.c
Kurt-von-Laven: Do you mean that `remove` and `purge` should not accept a `GError **`? The reason they currently do is that they can fail since they each perform writes to disk.

daemon/emer-daemon.c
```diff
@@ -211,15 +219,85 @@ is_uuid (GVariant *variant)
...
-  g_variant_unref (variant);
+  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
+
+  for (gint i = 0; i < num_events; i++)
+    {
+      GVariant *curr_variant = g_ptr_array_index (priv->variant_array, i);
+      priv->num_bytes_buffered -= get_size_with_type (curr_variant);
+    }
+
+  if (num_events > 0)
```
cosimoc: I don't think anything will happen if you call the function with zero

daemon/emer-daemon.c
```diff
@@ -211,15 +219,85 @@ is_uuid (GVariant *variant)
...
+              GVariant   *event)
+{
+  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
+
+  g_variant_ref_sink (event);
+  gsize size_with_type = get_size_with_type (event);
+  gsize new_bytes_buffered = priv->num_bytes_buffered + size_with_type;
+  if (new_bytes_buffered > priv->max_bytes_buffered ||
+      /* Don't get wedged by a large event. */
+      new_bytes_buffered > MAX_REQUEST_PAYLOAD)
```
cosimoc: Does this check not always restrict the total size regardless of what is set in `priv->max_bytes_buffered`? Maybe you intended to check `size_with_type > MAX_REQUEST_PAYLOAD`?

daemon/emer-daemon.c
```diff
@@ -211,15 +219,85 @@ is_uuid (GVariant *variant)
...
+    {
+      g_variant_unref (event);
```
cosimoc: Also wonder if we should not trigger a warning when dropping events on the floor here.

daemon/emer-daemon.c
Kurt-von-Laven: I was surprised by [this](https://github.com/GNOME/glib/blob/master/glib/garray.c#L1266), but `GPtrArray` explodes in your face if you try `g_ptr_array_remove_range (ptr_array, 0, 0)` on an empty array. I figured if we have to test for this case, we may as well do the test that saves the call to `g_ptr_array_remove_range` most often (as opposed to `priv->variant_array->len > 0`).

daemon/emer-daemon.c
Kurt-von-Laven: Yikes; good catch. Yes, that is what I intended.

daemon/emer-daemon.c
Kurt-von-Laven: Not sure; I worry it would flood the logs of users who are permanently offline if we logged every time the buffer was full. I could see having a warning for an egregiously large event, although it would be a lot more useful to have such a warning in the event recorder daemon where a programmer adding a new event might actually see it. What do you think?

cosimoc: Not sure I follow the last part: this is the event recorder daemon. But I get your point. I think it would be nice to only log the first time during the daemon's lifecycle an event is dropped because the buffer is full.

daemon/emer-daemon.c
cosimoc: Oh fair enough; in that case though you could move the check at the beginning of the function :)

daemon/emer-circular-file.c
cosimoc: Yeah; I see that is because of the disk writes, but I am thinking that it's odd that you need to manually save after adding data to the buffer but not when removing.

daemon/emer-circular-file.c
cosimoc: Couldn't you avoid writing the metadata file too when there are no bytes then?

daemon/emer-persistent-cache.c
```diff
@@ -501,121 +431,99 @@ compute_boot_offset (EmerPersistentCache *self,
...
-      g_warning ("Could not find a valid boot offset in the metadata file. "
-                 "Error: %s.", error->message);
-      g_error_free (error);
-      return reset_boot_offset_metadata_file (self, &relative_time,
-                                              &absolute_time);
+      g_propagate_prefixed_error (error, local_error,
+                                  "Failed to read boot offset from metadata "
+                                  "file %s. ",
+                                  priv->boot_metadata_file_path);
+      return FALSE;
```
cosimoc: No need to call `reset_boot_offset_metadata_file()` here anymore?

daemon/emer-persistent-cache.c
```diff
@@ -501,121 +431,99 @@ compute_boot_offset (EmerPersistentCache *self,
...
+      gchar system_boot_id_string[BOOT_ID_FILE_LENGTH];
+      uuid_unparse_lower (system_boot_id, system_boot_id_string);
+
+      gboolean was_reset = FALSE;
+      gboolean write_succeeded =
+        save_timing_metadata (self, &relative_time, &absolute_time,
+                              &boot_offset, system_boot_id_string, &was_reset,
+                              error);
+
+      if (!write_succeeded)
```
cosimoc: No need to call `reset_boot_offset_metadata_file()` here anymore?

daemon/emer-persistent-cache.c
```diff
@@ -1396,12 +591,10 @@ apply_cache_versioning (EmerPersistentCache *self,
 
   if (g_mkdir_with_parents (priv->cache_directory, 02774) != 0)
     {
-      const gchar *err_str = g_strerror (errno); // Don't free.
-      g_critical ("Failed to create directory: %s. Error: %s.",
-                  priv->cache_directory, err_str);
+      const gchar *error_string = g_strerror (errno); // Don't free.
```
cosimoc: No need for the extra comment here

daemon/emer-persistent-cache.c
```diff
@@ -1664,38 +868,210 @@ emer_persistent_cache_initable_init (GInitableIface *iface)
...
+    update_boot_offset (self, always_update_timestamps, error);
+  if (!update_succeeded)
+    return FALSE;
+
+  if (offset != NULL)
+    *offset = priv->boot_offset;
+
+  return TRUE;
+}
+
```
cosimoc: Extra empty line

daemon/emer-persistent-cache.c
```diff
@@ -1664,38 +868,210 @@ emer_persistent_cache_initable_init (GInitableIface *iface)
...
+      memcpy (elem + type_length, variant_data, variant_size);
+      g_variant_unref (normalized_variant);
+      if (!emer_circular_file_append (priv->variant_file, elem, elem_size))
+        break;
+    }
+
+  if (!emer_circular_file_save (priv->variant_file, error))
+    return FALSE;
+
+  for (gsize i = 0; i < curr_variants_stored; i++)
```
cosimoc: By looking at how this function is called from the daemon, is this loop really needed? As far as I can see, `remove_range()` will already unref all the variants in the stored range.

daemon/emer-persistent-cache.h
```diff
@@ -61,24 +61,7 @@ G_BEGIN_DECLS
...
-#define AGGREGATE_SUFFIX  "aggregate.metrics"
-#define SEQUENCE_SUFFIX   "sequence.metrics"
-
-/*
- * GVariant Types for deserializing metrics.
- */
-#define INDIVIDUAL_TYPE "(uayxmv)"
-#define AGGREGATE_TYPE  "(uayxxmv)"
-#define SEQUENCE_TYPE   "(uaya(xmv))"
+#define VARIANT_FILENAME "variants.dat"
```
cosimoc: dat variant

tests/daemon/mock-persistent-cache.c
```diff
@@ -115,99 +100,73 @@ emer_persistent_cache_get_boot_time_offset (EmerPersistentCache *self,
...
+    emer_persistent_cache_get_instance_private (self);
 
-static void
-append_aggregates (GArray         *aggregate_variants,
-                   AggregateEvent *aggregate_structs,
-                   gint            num_aggregates_buffered)
-{
-  for (gint i = 0; i < num_aggregates_buffered; i++)
+  gsize curr_num_variants = 0;
+  for (gsize curr_num_bytes = 0; curr_num_variants < priv->variant_array->len;
```
cosimoc: See other comment about a similar loop

daemon/emer-persistent-cache.h
Kurt-von-Laven: Having trouble interpreting this comment

cosimoc: No interpretation needed. Bad joke on my part.

daemon/emer-persistent-cache.c
Kurt-von-Laven: It isn't strictly speaking necessary, but properly handling the floating references makes the code easier to test and more robust to changes to the daemon in the future IMO.

daemon/emer-persistent-cache.c
Kurt-von-Laven: We can, but I don't see how we could get into this state unless the user is mucking with our files. If the user has indeed modified the boot metadata file, I'd rather we not receive metrics from that client at all (who knows if they would be valid?) as opposed to risking breaking something even worse in this difficult-to-test edge case.

daemon/emer-persistent-cache.c
Kurt-von-Laven: It seems pretty pointless in retrospect to react to a failure to write the boot metadata file by writing the boot metadata file.

daemon/emer-persistent-cache.c
cosimoc: My point is not about the floating references; I would not want the variants to be doubly unreffed. Is it responsibility of the caller to unref the removed range or not?

daemon/emer-persistent-cache.c
cosimoc: I don't disagree; just pointing out the previous code used to do that.

daemon/emer-persistent-cache.c
cosimoc: Yeah; just commented because the previous code did so.

tests/daemon/test-daemon.c
```diff
@@ -990,7 +957,7 @@ setup (Fixture      *fixture,
                           fixture->mock_network_send_provider,
                           fixture->mock_permissions_provider,
                           fixture->mock_persistent_cache,
-                          20); // Buffer length
+                          100000); // Max bytes buffered
```
cosimoc: Maybe change this comment to the other style you use above while you're touching this?

daemon/emer-daemon.c
Kurt-von-Laven: Oops; I meant to say it would probably be more useful to log the attempt to record an egregiously large event in the event recorder API (`eos-metrics/eosmetrics/emtr-event-recorder.c`).

That might complicate the test cases a bit, but maybe we could use an info level log assuming those don't trigger test failures?

cosimoc: Yeah, dropping events in the metrics public library sounds like a good idea. Not for this issue though.
You can use `g_test_expect_message()` to make tests robust to failure messages or test the messages themselves.

tests/daemon/test-persistent-cache.c
```diff
@@ -349,1064 +302,449 @@ boot_timestamp_is_valid (gint64 previous_relative_time,
...
-assert_aggregates_equal_variants (AggregateEvent *aggregate_array,
-                                  gint            aggregate_array_length,
-                                  GVariant      **variants)
-{
-  for (gint i = 0; i < aggregate_array_length || variants[i] != NULL; i++)
-    {
-      g_assert_true (i < aggregate_array_length && variants[i] != NULL);
+  /* These arbitrary values should be overwritten. */
+  gsize num_variants_read = num_variants + 1;
+  guint64 token = total_size_when_stored + 1;
```
cosimoc: Why even initialize these variables then?

tests/daemon/test-persistent-cache.c
```diff
@@ -349,1064 +302,449 @@ boot_timestamp_is_valid (gint64 previous_relative_time,
...
-  gboolean success =
-    store_many (cache, &num_singulars_made, &num_aggregates_made,
-                &num_sequences_made, &num_singulars_stored,
-                &num_aggregates_stored, &num_sequences_stored, &capacity);
-  g_object_unref (cache);
-  g_assert_true (success);
+  gsize num_variants;
+  GVariant **variants = make_many_variants (&num_variants);
+  assert_variants_stored (cache, variants, num_variants);
+  g_free (variants);
```
cosimoc: Other tests destroy the variants after calling `assert_variants_stored()`

tests/daemon/test-persistent-cache.c
```diff
@@ -349,1064 +302,449 @@ boot_timestamp_is_valid (gint64 previous_relative_time,
...
-  make_singular_event (0, singular_array);
-
-  AggregateEvent aggregate_array[1];
-  make_aggregate_event (0, aggregate_array);
-
-  SequenceEvent sequence_array[1];
-  make_sequence_event (0, sequence_array);
+  GVariant *variants[] = { make_variant (0) };
+  gsize num_variants = G_N_ELEMENTS (variants);
+  assert_variants_stored (cache, variants, num_variants);
```
cosimoc: Probably need to unref the one variant afterwards here.

tests/daemon/test-persistent-cache.c
Kurt-von-Laven: It makes the tests more deterministic. In particular, zero is the most common value in uninitialized memory, and often these values are supposed to be zero.

daemon/emer-persistent-cache.c
Kurt-von-Laven: Yeah, this change belongs in a separate commit along with the one below. I didn't catch on to this when I reviewed the previous code originally.

daemon/emer-persistent-cache.c
Kurt-von-Laven: The daemon ref sinks its variants, so I don't see where these would be doubly unreffed. The caller should unref the variants if they took a reference on them but not if they didn't. I believe this is the standard behavior of most APIs in glib.

cosimoc:
@Kurt-von-Laven Forgot to comment here at the end of the day, but I'm actually done with the first round of CR. Approach looks good to me and code too.
Looking forward to see the finished version of this PR!

daemon/emer-circular-file.c
Kurt-von-Laven: I don't think so. Presumably we aren't supposed to leave an arbitrarily large data file on disk if we resize to a smaller max size. It also seems counter-intuitive to me that if all of the following operations were successful, one would be left with an `EmerCircularFile` that still contained `bar`. We can't overwrite the data file without writing the metadata file either because that could result in the two being left in an inconsistent state.

```
EmerCircularFile *circular_file = emer_circular_file_new ("foo", 10, NULL);
emer_circular_file_append (circular_file, bar, 5);
g_object_unref (circular_file);
circular_file = emer_circular_file_new ("foo", 0, NULL); /* Write nothing because num_bytes == 0. */
g_object_unref (circular_file);
circular_file = emer_circular_file_new ("foo", 20, NULL);
```

daemon/emer-circular-file.c
Kurt-von-Laven: Yeah, that does look nicer.

daemon/emer-circular-file.c
Kurt-von-Laven: Good idea.

daemon/emer-circular-file.c
Kurt-von-Laven: I not really a big fan of `may_fail_init` either, but the function this line is in is already named `emer_circular_file_initable_init`. Other suggestions?

cosimoc: Yeah; you can rename this to `emer_circular_file_initable_iface_init`

daemon/emer-circular-file.c
Kurt-von-Laven: `append` and `save` are separate so that we can avoid writing each variant to disk individually when we flush to the persistent cache every hour. Roddy felt pretty strongly that there should be as little writing to disk as possible in the metrics system so that it doesn't wear out the SD card or eMMC storage. I don't think I understand what you have in mind though.

daemon/emer-circular-file.c
Kurt-von-Laven: That will underflow though :(.

cosimoc: Not sure I understand. Can you make `excess_data` large enough that it does not underflow?

Kurt-von-Laven: The reason it can underflow is that `gsize` is an unsigned type. I could use a `gint64`, which is technically the wrong type for the job, but wouldn't cause a problem until people had exabytes of RAM :P.

daemon/emer-circular-file.c
Kurt-von-Laven: This being the function currently known as `emer_circular_file_may_fail_init` or `emer_circular_file_initable_init`?

daemon/emer-circular-file.c
cosimoc: Or a `gssize`? Anyway yeah, sounds better to use a type that avoids that.

daemon/emer-circular-file.c
cosimoc: `emer_circular_file_initable_init` -> `emer_cicrular_file_initable_iface_init`
`emer_circular_file_may_fail_init` -> `emer_circular_file_initable_init`

daemon/emer-circular-file.c
Kurt-von-Laven: `gssize` is only guaranteed to go down to `-1` (http://pubs.opengroup.org/onlinepubs/009696699/basedefs/sys/types.h.html), but I suspect it would work in practice.

daemon/emer-daemon.c
Kurt-von-Laven: Good point!

daemon/emer-daemon.c
Kurt-von-Laven: It's the same value, but it is just a coincidence that we agreed the metrics system can use 100 kB of RAM and upload at most 100 kB at a time.

tests/daemon/test-persistent-cache.c
Kurt-von-Laven: That actually causes a crash because `assert_variants_stored` consumes the floating reference.

tests/daemon/test-persistent-cache.c
Kurt-von-Laven: Only the tests that call `g_variant_ref_sink` should be unreffing the variants.

PR #195, opened 2015-07-30
Corrected log message and comment.

Kurt-von-Laven:
Events are no longer dropped when network requests fail.

[endlessm/eos-sdk#3439]

### DIFF
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -52,7 +52,7 @@
 
 /*
  * The maximum number of attempts to upload a particular batch of metric events
- * before giving up and dropping them.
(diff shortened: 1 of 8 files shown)


Kurt-von-Laven:
https://github.com/endlessm/eos-event-recorder-daemon/commit/c36daf4ec1344200ebadf8ddd7ffe11f8bfe6ba6 is the commit that I feel should be cherry-picked to eos2.4 because I don't want the log message confusing folks within Endless. The others are pretty benign in my opinion given that the repo still builds and all the tests still pass, so merging the entire master branch to eos2.4 seems reasonable too.

PR #196, opened 2015-07-30
Perform full upload when explicitly requested.

Kurt-von-Laven:
As documented in the D-Bus interface of the event recorder daemon, an
explicitly requested (as opposed to timer-driven) upload should include
all available events, both in the persistent cache and in memory. We now
only limit the upload size to 100 kB for timer-driven uploads.

[endlessm/eos-sdk#3151]


Kurt-von-Laven:
Both of these commits should be cherry-picked to eos2.4 to avoid the complexity of maintaining different histories since the latter commit won't apply correctly without the former.

PR #197, opened 2015-08-06
Remove requirement on dbus.service

dbnicholson:
Since the service is type dbus, systemd will already add an implicit
dependency on dbus.socket (see systemd.service(5)). This allows the
event-recorder to begin initialization and then block in the dbus call
until the dbus service has taken over the socket from systemd.

This is how all other dbus service units on the system are managed and
allows systemd to start more services in parallel.

[endlessm/eos-shell#5502]


Kurt-von-Laven:
Good to know!

PR #200, opened 2015-08-12
Create cache size file at runtime.

ptomato:
Don't create it on make install since it's needed by one of our tests
and make check shouldn't assume that make install has been run.

[endlessm/eos-sdk#3151]


ptomato:
**CR pass**

PR #201, opened 2015-08-12
Support persistent cache dir command line arg.

Kurt-von-Laven:
Use the new argument in test-opt-out-integration.py to put the
persistent cache directory in a temporary directory so that
make check no longer assumes make install has been run.

[endlessm/eos-sdk#3151]


cosimoc:
This PR is against the wrong branch (master instead of eos2.4); I'll just cherry-pick these manually.

Kurt-von-Laven:
For the record, this was intentional since Philip was keeping the non-essential commits off master temporarily and then merging master to eos2.4 wholesale. Thanks for sorting it out.

PR #205, opened 2015-09-01
Free metrics from its inhibitions

wgreenberg:
We don't actually need to inhibit shutdown just to flush events to the
persistent cache, since systemd will provide us with a 90 second timeout
between SIGTERM and SIGKILL.

[endlessm/eos-sdk#3097]

### DIFF
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -125,7 +125,6 @@ typedef struct _NetworkCallbackData
 
 typedef struct _EmerDaemonPrivate
 {
-  gint shutdown_inhibitor;
   GDBusProxy *login_manager_proxy;
 
   guint network_send_interval;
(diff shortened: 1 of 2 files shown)


daemon/emer-daemon.c
```diff
@@ -1005,18 +933,8 @@ handle_login_manager_signal (GDBusProxy *dbus_proxy,
...
-        {
-          flush_to_persistent_cache (self);
-          update_timestamps (self);
-          release_shutdown_inhibitor (self);
-        }
-      else
-        {
-          inhibit_shutdown (self);
-        }
+      flush_to_persistent_cache (self);
```
Kurt-von-Laven: @wgreenberg @cosimoc is it intentional that we now flush to the persistent cache and update timestamps both when the system is shutting down and also when another process cancels shutdown? I was assuming we would only want to flush when the system is shutting down to save disk I/O on eMMC devices and SD cards.

Kurt-von-Laven: On second thought can't we remove `handle_login_manager_signal` entirely?

cosimoc: I think the chance a significant amount of events was recorded between those signals is very small - so small that it's not worth complicating the code.
In any case I would not oppose keeping the logic that only does this when `shutting_down` is TRUE

PR #206, opened 2015-09-02
Remove the shutdown hooks for timestamp updates/cache flushing

wgreenberg:
Accordingly, the unit tests for these were changed to assert that the
cache is simply flushed on when the daemon is finalized.

[endlessm/eos-sdk#3097]

### DIFF
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -68,12 +68,6 @@
 
 #define DEFAULT_NETWORK_SEND_FILENAME "network_send_file"
 
-#define WHAT "shutdown"
-#define WHO "EndlessOS Event Recorder Daemon"
-#define WHY "Flushing events to disk"
-#define MODE "delay"
-#define INHIBIT_ARGS "('" WHAT "', '" WHO "', '" WHY "', '" MODE "')"
-
 #define EVENT_VALUE_TYPE_STRING "(xmv)"
 #define EVENT_VALUE_ARRAY_TYPE_STRING "a" EVENT_VALUE_TYPE_STRING
 #define EVENT_VALUE_ARRAY_TYPE G_VARIANT_TYPE (EVENT_VALUE_ARRAY_TYPE_STRING)
@@ -125,8 +119,6 @@ typedef struct _NetworkCallbackData
 
 typedef struct _EmerDaemonPrivate
 {
-  GDBusProxy *login_manager_proxy;
-
   guint network_send_interval;
   GQueue *upload_queue;
   gboolean uploading;
@@ -910,94 +902,6 @@ handle_upload_finished (EmerDaemon *self)
 }
 
 static void
-update_timestamps (EmerDaemon *self)
-{
-  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
-
-  GError *error = NULL;
-  if (!emer_persistent_cache_get_boot_time_offset (priv->persistent_cache,
-                                                   NULL, TRUE, &error))
-    {
-      g_warning ("Persistent cache could not update timestamps: %s.",
-                 error->message);
-      g_error_free (error);
-    }
-}
-
-static void
-handle_login_manager_signal (GDBusProxy *dbus_proxy,
-                             gchar      *sender_name,
-                             gchar      *signal_name,
-                             GVariant   *parameters,
-                             EmerDaemon *self)
-{
-  if (g_strcmp0 ("PrepareForShutdown", signal_name) == 0)
-    {
-      flush_to_persistent_cache (self);
-      update_timestamps (self);
-    }
-}
-
-static void
-register_with_login_manager (EmerDaemon *self)
-{
-  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
-
-  g_signal_connect (priv->login_manager_proxy, "g-signal",
-                    G_CALLBACK (handle_login_manager_signal), self);
-}
-
-static gboolean
-has_owner (GDBusProxy *dbus_proxy)
-{
-  gchar *name_owner =
-    g_dbus_proxy_get_name_owner (dbus_proxy);
-  if (name_owner != NULL)
-    g_free (name_owner);
-
-  return name_owner != NULL;
-}
-
-static void
-handle_login_manager_name_owner_set (GDBusProxy *dbus_proxy,
-                                     GParamSpec *pspec,
-                                     EmerDaemon *self)
-{
-  if (has_owner (dbus_proxy))
-    register_with_login_manager (self);
-}
-
-static void
-connect_to_login_manager (EmerDaemon *self)
-{
-  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
-
-  GError *error = NULL;
-  priv->login_manager_proxy =
-    g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
-                                   G_DBUS_PROXY_FLAGS_DO_NOT_AUTO_START,
-                                   NULL /* GDBusInterfaceInfo */,
-                                   "org.freedesktop.login1",
-                                   "/org/freedesktop/login1",
-                                   "org.freedesktop.login1.Manager",
-                                   NULL /* GCancellable */,
-                                   &error);
-  if (priv->login_manager_proxy == NULL)
-    {
-      g_warning ("Error creating login manager D-Bus proxy: %s.",
-                 error->message);
-      g_error_free (error);
-      return;
-    }
-
-  if (has_owner (priv->login_manager_proxy))
-    register_with_login_manager (self);
-  else
-    g_signal_connect (priv->login_manager_proxy, "notify::g-name-owner",
-                      G_CALLBACK (handle_login_manager_name_owner_set), self);
-}
-
-static void
 on_permissions_changed (EmerPermissionsProvider *permissions_provider,
                         GParamSpec              *pspec,
                         EmerDaemon              *self)
@@ -1208,8 +1112,6 @@ emer_daemon_finalize (GObject *object)
   flush_to_persistent_cache (self);
   g_clear_object (&priv->persistent_cache);
 
-  g_clear_object (&priv->login_manager_proxy);
-
   g_queue_free_full (priv->upload_queue, g_object_unref);
 
   soup_session_abort (priv->http_session);
@@ -1380,8 +1282,6 @@ static void
 emer_daemon_init (EmerDaemon *self)
 {
   EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
-
-  connect_to_login_manager (self);
 
   priv->upload_queue = g_queue_new ();
 
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -368,7 +368,7 @@ get_offset_timestamps (EmerDaemon *self,
 
   gint64 boot_offset;
   if (!emer_persistent_cache_get_boot_time_offset (priv->persistent_cache,
-                                                   &boot_offset, FALSE, error))
+                                                   &boot_offset, error))
     return FALSE;
 
   *rel_timestamp_ptr += boot_offset;
@@ -1380,7 +1380,7 @@ emer_daemon_record_singular_event (EmerDaemon *self,
   gint64 boot_offset;
   GError *error = NULL;
   if (!emer_persistent_cache_get_boot_time_offset (priv->persistent_cache,
-                                                   &boot_offset, FALSE, &error))
+                                                   &boot_offset, &error))
     {
       g_warning ("Unable to correct event's relative timestamp. Dropping "
                  "event. Error: %s.", error->message);
@@ -1420,7 +1420,7 @@ emer_daemon_record_aggregate_event (EmerDaemon *self,
   gint64 boot_offset;
   GError *error = NULL;
   if (!emer_persistent_cache_get_boot_time_offset (priv->persistent_cache,
-                                                   &boot_offset, FALSE, &error))
+                                                   &boot_offset, &error))
     {
       g_warning ("Unable to correct event's relative timestamp. Dropping "
                  "event. Error: %s.", error->message);
@@ -1457,7 +1457,7 @@ emer_daemon_record_event_sequence (EmerDaemon *self,
   gint64 boot_offset;
   GError *error = NULL;
   if (!emer_persistent_cache_get_boot_time_offset (priv->persistent_cache,
-                                                   &boot_offset, FALSE, &error))
+                                                   &boot_offset, &error))
     {
       g_warning ("Unable to correct event's relative timestamp. Dropping "
                  "event. Error: %s.", error->message);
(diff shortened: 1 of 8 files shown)


daemon/emer-daemon.c
```diff
@@ -959,15 +921,6 @@ has_owner (GDBusProxy *dbus_proxy)
...
-handle_login_manager_name_owner_set (GDBusProxy *dbus_proxy,
-                                     GParamSpec *pspec,
-                                     EmerDaemon *self)
-{
-  if (has_owner (dbus_proxy))
-    register_with_login_manager (self);
-}
-
-static void
 connect_to_login_manager (EmerDaemon *self)
```
Kurt-von-Laven: We should remove this entire method now as well as removing `login_manager_proxy` from private state.

tests/daemon/test-daemon.c
```diff
@@ -185,9 +179,6 @@ process_logind_line (GString *line,
 {
```
Kurt-von-Laven: We should be able to remove this entire method now.

tests/daemon/test-daemon.c
```diff
@@ -1260,13 +1205,14 @@ test_daemon_flushes_to_persistent_cache_on_shutdown (Fixture      *fixture,
...
 
   record_singulars (fixture->test_object);
-  emit_shutdown_signal (TRUE);
 
   /* Wait for daemon to handle the signal. */
   while (g_main_context_pending (NULL))
     g_main_context_iteration (NULL, TRUE);
 
-  terminate_subprocess_and_wait (fixture->logind_mock);
+  // unref the daemon, causing it to finalize
```
Kurt-von-Laven: nit: Prefer `/* Comments like this */` per a dev architecture style discussion.

Kurt-von-Laven: nit: "unref" --> "Unref"

Kurt-von-Laven: nit: Terminal period.

tests/daemon/test-daemon.c
```diff
@@ -1260,13 +1205,14 @@ test_daemon_flushes_to_persistent_cache_on_shutdown (Fixture      *fixture,
...
+  g_object_unref (fixture->test_object);
+  fixture->test_object = NULL;
```
Kurt-von-Laven: nit: Prefer `g_clear_object (&fixture->test_object)` to this pattern.

tests/daemon/test-daemon.c
```diff
@@ -1015,7 +984,8 @@ static void
 teardown (Fixture      *fixture,
           gconstpointer unused)
 {
-  g_object_unref (fixture->test_object);
+  if (fixture->test_object != NULL)
+    g_object_unref (fixture->test_object);
```
Kurt-von-Laven: nit: Prefer `g_clear_object (&fixture->test_object)` to this pattern.

tests/daemon/test-daemon.c
```diff
@@ -63,12 +63,6 @@
```
Kurt-von-Laven: nit: Can remove MEANINGLESS_EVENT_2 (pre-existing defect).

tests/daemon/test-daemon.c
```diff
@@ -1225,32 +1195,7 @@ test_daemon_does_not_record_sequences_when_daemon_disabled (Fixture      *fixtur
...
-
-  terminate_subprocess_and_wait (fixture->logind_mock);
-}
-
-static void
-test_daemon_flushes_to_persistent_cache_on_shutdown (Fixture      *fixture,
+test_daemon_flushes_to_persistent_cache_on_finalize (Fixture      *fixture,
                                                      gconstpointer unused)
 {
   start_mock_logind_service (fixture);
```
Kurt-von-Laven: rm

tests/daemon/test-daemon.c
```diff
@@ -1260,13 +1205,14 @@ test_daemon_flushes_to_persistent_cache_on_shutdown (Fixture      *fixture,
                           NULL /* user_data */);
 
   record_singulars (fixture->test_object);
-  emit_shutdown_signal (TRUE);
 
   /* Wait for daemon to handle the signal. */
   while (g_main_context_pending (NULL))
```
Kurt-von-Laven: I believe we can remove this atrocious while loop now.

tests/daemon/test-daemon.c
```diff
@@ -63,12 +63,6 @@
...
 
-#define EXPECTED_INHIBIT_SHUTDOWN_ARGS \
-  "\"shutdown\" " \
-  "\"EndlessOS Event Recorder Daemon\" " \
-  "\"Flushing events to disk\" " \
-  "\"delay\""
-
 typedef struct _Fixture
 {
   EmerDaemon *test_object;
```
Kurt-von-Laven: We should be able to remove `logind_mock` from the fixture.

Kurt-von-Laven:
Revise ready!

daemon/emer-daemon.c
wgreenberg: ah, since we're no longer grabbing an inhibitor lock?

tests/daemon/test-daemon.c
wgreenberg: This comment got orphaned since it wasn't on a diff line, but I removed the logind methods

tests/daemon/test-daemon.c
wgreenberg: Done dealio

daemon/emer-daemon.c
Kurt-von-Laven: Exactly. That was the only thing we were using the login manager for in this class.

daemon/emer-daemon.c
```diff
@@ -125,8 +125,6 @@ typedef struct _NetworkCallbackData
```
Kurt-von-Laven: Sorry I missed this the first time around, but we can also rm `INHIBIT_ARGS` and friends ^^.

daemon/emer-daemon.c
```diff
@@ -910,94 +908,6 @@ handle_upload_finished (EmerDaemon *self)
 }
 
 static void
-update_timestamps (EmerDaemon *self)
-{
-  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
-
-  GError *error = NULL;
-  if (!emer_persistent_cache_get_boot_time_offset (priv->persistent_cache,
```
Kurt-von-Laven: Since this was the only call to `emer_persistent_cache_get_boot_time_offset` that passed `TRUE` for the `always_update_timestamps` parameter, we should also remove the `always_update_timestamps` parameter from `emer_persistent_cache_get_boot_time_offset`. This will discourage someone from reintroducing a call that passes `TRUE` to `always_update_timestamps`, which was something of a hack in the first place.

tests/daemon/test-daemon.c
```diff
@@ -1225,48 +1175,13 @@ test_daemon_does_not_record_sequences_when_daemon_disabled (Fixture      *fixtur
 }
 
 static void
-test_daemon_updates_timestamps_on_shutdown (Fixture      *fixture,
```
Kurt-von-Laven: Upon further reflection, the moral equivalent of this test would be that the persistent cache gets released at finalize time.

tests/test-opt-out-integration.py
```diff
@@ -47,8 +47,7 @@ def setUpClass(klass):
     def setUp(self):
         """Start the event recorder on the mock system bus."""
 
-        # Put polkitd and logind mocks onto the mock system bus
-        (self.login_popen, self.login_obj) = self.spawn_server_template('logind')
+        # Put polkitd mocks onto the mock system bus
```
Kurt-von-Laven: nit: Terminal period (pre-existing defect).

Kurt-von-Laven:
Revise ready.

daemon/emer-daemon.c
wgreenberg: There's a number of tests in `test-persistent-cache.c` which pass TRUE for that method. Are you sure they should all be removed as well?

tests/daemon/test-daemon.c
wgreenberg: IMO we just remove the test then, it seems superfluous to add unit tests for memory leaks

Kurt-von-Laven: I agree with that general statement, but it's not the memory management that we're interested in testing here; it's the accuracy of the timestamps, which is improved when the persistent cache is properly finalized. The fact that the update timestamps call happens during finalization is purely a coincidence in moral terms.

wgreenberg: Not sure I follow. Are you saying the test should be added with the same `finalize` modification as the flushing test?

tests/daemon/test-daemon.c
wgreenberg: done!

daemon/emer-daemon.c
Kurt-von-Laven: Sorry, I meant to specify limit the scope of my sweeping claim to production code. In general, those tests shouldn't be removed. They can all be refactored to work without an `always_update_timestamps` parameter though. In some cases, the value of the parameter was irrelevant to the logic being tested in the first place. In some cases, the follow-on logic that tests whether the timestamps have been updated will need to be removed or modified. In some cases, we may need to replace release and recreate the persistent cache in place of the call to `emer_persistent_cache_get_boot_time_offset (/* ... */, TRUE, /* ... */)`. If you have doubts about specific test cases, please don't hesitate to ask. Devin put a considerable amount of effort into documenting them, but the fact that they required so much documentation in the first place suggests that they aren't all as straightforward as we would like.

tests/daemon/test-daemon.c
Kurt-von-Laven: What I'm trying to say is that we should add a test that does nothing more and nothing less than verify that calling `g_clear_object (&fixture->test_object)` causes `fixture->mock_persistent_cache` to be unreffed. The underlying reason for this test is that leaking the persistent cache would have more significant consequences than the standard waste of memory; it's actually a timestamp data quality concern. Does that make sense?

wgreenberg:
@Kurt-von-Laven review ready again!

daemon/emer-persistent-cache.c
```diff
@@ -966,17 +966,13 @@ emer_persistent_cache_cost (GVariant *variant)
 gboolean
```
Kurt-von-Laven: nit: The function comment still refers to `always_update_timestamps`.

tests/daemon/test-persistent-cache.c
```diff
@@ -878,24 +878,13 @@ static void
...
-  get_succeeded =
-    emer_persistent_cache_get_boot_time_offset (cache, NULL, TRUE, &error);
-  g_assert_no_error (error);
-  g_assert_true (get_succeeded);
+  EmerPersistentCache *cache = make_testing_cache ();
+
+  gint64 first_offset = read_offset ();
 
   g_assert_true (boot_timestamp_is_valid (relative_time, absolute_time));
   gint64 second_offset = read_offset();
```
Kurt-von-Laven: We should remove this line and replace `second_offset` with `0` below. The way this code is written now makes it seem like the `g_assert_cmpint` is testing that `boot_timestamp_is_valid` didn't change the boot offset, but `boot_timestamp_is_valid` is a testing function, so even if it could have such a bug, it wouldn't be worth testing for it.

tests/daemon/test-persistent-cache.c
```diff
@@ -969,7 +958,7 @@ test_persistent_cache_does_not_compute_offset_when_boot_id_is_same (gboolean
```
Kurt-von-Laven: The function comment has fallen out of date with the implementation.

tests/daemon/test-persistent-cache.c
```diff
@@ -987,7 +976,7 @@ test_persistent_cache_does_not_compute_offset_when_boot_id_is_same (gboolean
 
   // This call should have to compute the boot offset itself.
   get_succeeded =
-    emer_persistent_cache_get_boot_time_offset (cache2, NULL, TRUE, &error);
+    emer_persistent_cache_get_boot_time_offset (cache2, NULL, &error);
```
Kurt-von-Laven: This call can be removed, and the comment above can be moved to describe `EmerPersistentCache *cache2 = make_testing_cache ();` (pre-existing defect).

tests/daemon/test-persistent-cache.c
```diff
@@ -969,7 +958,7 @@ test_persistent_cache_does_not_compute_offset_when_boot_id_is_same (gboolean
 
   GError *error = NULL;
   gboolean get_succeeded =
-    emer_persistent_cache_get_boot_time_offset (cache, NULL, TRUE, &error);
+    emer_persistent_cache_get_boot_time_offset (cache, NULL, &error);
```
Kurt-von-Laven: This call can be removed (pre-existing defect).

tests/daemon/test-persistent-cache.c
```diff
@@ -1002,7 +991,7 @@ test_persistent_cache_does_not_compute_offset_when_boot_id_is_same (gboolean
   EmerPersistentCache *cache3 = make_testing_cache ();
 
   get_succeeded =
-    emer_persistent_cache_get_boot_time_offset (cache3, NULL, TRUE, &error);
+    emer_persistent_cache_get_boot_time_offset (cache3, NULL, &error);
```
Kurt-von-Laven: This call can be removed (pre-existing defect).

tests/daemon/test-persistent-cache.c
```diff
@@ -1081,7 +1060,7 @@ test_persistent_cache_get_offset_wont_update_timestamps_if_it_isnt_supposed_to (
```
Kurt-von-Laven: The documentation for this method has fallen out of sync with the implementation.

Kurt-von-Laven: Alternatively, you could remove this test case entirely at your discretion since it's effectively testing the functionality of a parameter that has been removed.

tests/daemon/test-persistent-cache.c
```diff
@@ -1138,7 +1086,7 @@ test_persistent_cache_updates_timestamps_on_finalize (gboolean     *unused,
 
   // Update metadata file to reasonable values.
   gboolean get_succeeded =
-    emer_persistent_cache_get_boot_time_offset (cache, NULL, TRUE, &error);
+    emer_persistent_cache_get_boot_time_offset (cache, NULL, &error);
```
Kurt-von-Laven: Can remove this call entirely (pre-existing defect).

tests/daemon/test-persistent-cache.c
```diff
@@ -1179,7 +1127,7 @@ test_persistent_cache_get_offset_can_build_boot_metadata_file (gboolean     *unu
    * always_update_timestamps parameter is FALSE.
```
Kurt-von-Laven: Comment refers to removed symbol.

Kurt-von-Laven: The comment above that says "Don't write a default boot offset file [...]" should be replaced with:
```g_assert_cmpint (g_unlink (TEST_DIRECTORY BOOT_OFFSET_METADATA_FILE), ==, 0);``` (pre-existing defect).

tests/daemon/test-daemon.c
Kurt-von-Laven: @wgreenberg Did you omit the test case discussed here intentionally?

Kurt-von-Laven:
Revise ready!

PR #207, opened 2015-09-03
Sdk/3097

wgreenberg:
[endlessm/eos-sdk#3097]

### DIFF
--- a/daemon/emer-persistent-cache.c
+++ b/daemon/emer-persistent-cache.c
@@ -953,12 +953,8 @@ emer_persistent_cache_cost (GVariant *variant)
   return type_string_length + variant_size;
 }
 
-/* Gets the boot time offset and stores it in the out parameter offset.
- * If the always_update_timestamps parameter is FALSE, does not write to disk
- * solely to update the timestamps during this operation unless the boot id is
- * out of date or some corruption is detected that prompts a total rewrite of
- * the boot timing metadata file. Pass an offset of NULL if you don't care about
- * its value.
+/* Gets the boot time offset and stores it in the out parameter offset. Pass
+ * an offset of NULL if you don't care about its value.
  *
  * Returns TRUE on success. Returns FALSE on failure and sets the error unless
  * it's NULL. Does not alter the offset out parameter on failure.
--- a/daemon/emer-persistent-cache.c
+++ b/daemon/emer-persistent-cache.c
@@ -778,8 +778,11 @@ emer_persistent_cache_finalize (GObject *object)
       g_error_free (error);
     }
 
-  g_clear_pointer (&priv->boot_offset_update_timeout_source_id,
-                   g_source_remove);
+  if (priv->boot_offset_update_timeout_source_id != 0)
+    {
+      g_source_remove (priv->boot_offset_update_timeout_source_id);
+      priv->boot_offset_update_timeout_source_id = 0;
+    }
 
   g_clear_object (&priv->cache_size_provider);
   g_clear_object (&priv->boot_id_provider);
--- a/tests/daemon/test-persistent-cache.c
+++ b/tests/daemon/test-persistent-cache.c
@@ -887,8 +887,6 @@ test_persistent_cache_builds_boot_metadata_file (gboolean     *unused,
   gint64 offset = read_offset ();
 
   g_assert_true (boot_timestamp_is_valid (relative_time, absolute_time));
-
-  // The offset should not have changed.
   g_assert_cmpint (offset, ==, 0);
   g_assert_true (boot_offset_was_reset ());
 
@@ -939,15 +937,12 @@ test_persistent_cache_computes_reasonable_offset (gboolean     *unused,
  * Tests that releasing and recreating the persistent cache within the same boot
  * doesn't change the boot offset.
  *
- * Triggers the computation of a new boot offset by asking for the new boot
- * offset with no preexisting boot metadata file, which triggers a reset to
- * offset 0 and the current boot id. Then unrefs the persistent cache and makes
- * it anew, clearing its in-memory cache. Then mutates the metadata file to
- * simulate a new boot. Then requests the boot offset again to prompt the
- * persistent cache to compute a new offset. Then unrefs and recreates the
- * persistent cache again to clear its in-memory cache. Finally, requests the
- * boot offset, prompting the persistent cache to write new timestamps without
- * recomputing the boot offset.
+ * Instantiates a persistent cache with no preexisting boot metadata file,
+ * which triggers a reset to a boot offset of 0. Then unrefs the persistent
+ * cache, mutates the boot ID in the metadata file, and makes the persistent
+ * cache anew to simulate a new boot and trigger computation of a non-zero boot
+ * offset. Finally, unrefs and makes anew the persistent cache once more,
+ * prompting it to detect the unchanged boot ID and preserve the boot offset.
  */
 static void
 test_persistent_cache_does_not_compute_offset_when_boot_id_is_same (gboolean     *unused,
@@ -965,6 +960,7 @@ test_persistent_cache_does_not_compute_offset_when_boot_id_is_same (gboolean    
   gint64 absolute_time;
   g_assert_true (emtr_util_get_current_time (CLOCK_REALTIME, &absolute_time));
 
+  // This call should have to compute the boot offset itself.
   EmerPersistentCache *cache2 = make_testing_cache ();
 
   g_assert_true (boot_timestamp_is_valid (relative_time, absolute_time));
--- a/tests/daemon/test-persistent-cache.c
+++ b/tests/daemon/test-persistent-cache.c
@@ -884,13 +884,12 @@ test_persistent_cache_builds_boot_metadata_file (gboolean     *unused,
 
   EmerPersistentCache *cache = make_testing_cache ();
 
-  gint64 first_offset = read_offset ();
+  gint64 offset = read_offset ();
 
   g_assert_true (boot_timestamp_is_valid (relative_time, absolute_time));
-  gint64 second_offset = read_offset();
 
   // The offset should not have changed.
-  g_assert_cmpint (first_offset, ==, second_offset);
+  g_assert_cmpint (offset, ==, 0);
   g_assert_true (boot_offset_was_reset ());
 
   g_object_unref (cache);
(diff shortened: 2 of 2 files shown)


tests/daemon/test-persistent-cache.c
```diff
@@ -954,33 +953,20 @@ static void
...
-  EmerPersistentCache *cache = make_testing_cache ();
-
-  GError *error = NULL;
-  gboolean get_succeeded =
-    emer_persistent_cache_get_boot_time_offset (cache, NULL, &error);
-  g_assert_no_error (error);
-  g_assert_true (get_succeeded);
-  g_assert_true (boot_offset_was_reset ());
-
   gint64 relative_time;
```
Kurt-von-Laven: We should probably wait until just before the second call to `make_testing_cache` to get these timestamps so that we confine the the timestamps in the metadata file to the narrowest possible range. This should benefit both readability and testing fidelity.

tests/daemon/test-persistent-cache.c
```diff
@@ -954,33 +953,20 @@ static void
...
+  EmerPersistentCache *cache = make_testing_cache ();
+
+  g_assert_true (boot_offset_was_reset ());
+
   g_object_unref (cache);
   set_boot_id_in_metadata_file (FAKE_BOOT_ID);
 
   EmerPersistentCache *cache2 = make_testing_cache ();
 
-  // This call should have to compute the boot offset itself.
```
Kurt-von-Laven: This comment can be moved to describe the second `make_testing_cache` call rather than be deleted.

tests/daemon/test-persistent-cache.c
```diff
@@ -954,33 +953,20 @@ static void
 test_persistent_cache_does_not_compute_offset_when_boot_id_is_same (gboolean     *unused,
```
Kurt-von-Laven: Here is my attempt to clarify/correct the documentation for this method. If it doesn't help you understand how this function works, then it clearly isn't doing a very good job.

```
 /* Tests that releasing and recreating the persistent cache within the same boot
  * doesn't change the boot offset.
  *
  * Instantiates a persistent cache with no preexisting boot metadata file,
  * which triggers a reset to a boot offset of 0. Then unrefs the persistent
  * cache, mutates the boot ID in the metadata file, and makes the persistent
  * cache anew to simulate a new boot and trigger computation of a non-zero boot
  * offset. Finally, unrefs and makes anew the persistent cache once more,
  * prompting it to detect the unchanged boot ID and preserve the boot offset.
  */
```

tests/daemon/test-persistent-cache.c
```diff
@@ -1121,11 +1063,6 @@ test_persistent_cache_get_offset_can_build_boot_metadata_file (gboolean     *unu
    */
```
Kurt-von-Laven: Not sure if you didn't see this comment last time or chose to leave it for a separate issue:

The comment above that says "Don't write a default boot offset file [...]" should be replaced with:
```g_assert_cmpint (g_unlink (TEST_DIRECTORY BOOT_OFFSET_METADATA_FILE), ==, 0);``` (pre-existing defect).

File: tests/daemon/test-persistent-cache.c [hunk 5]
```diff
@@ -884,13 +884,12 @@ test_persistent_cache_builds_boot_metadata_file (gboolean     *unused,
 
   EmerPersistentCache *cache = make_testing_cache ();
 
-  gint64 first_offset = read_offset ();
+  gint64 offset = read_offset ();
 
   g_assert_true (boot_timestamp_is_valid (relative_time, absolute_time));
-  gint64 second_offset = read_offset();
 
   // The offset should not have changed.
```
Kurt-von-Laven: This comment has now fallen out of sync with the implementation. We can probably just remove it.

Kurt-von-Laven:
Looking pretty righteous at this point. Revise ready!

daemon/emer-persistent-cache.c
```diff
@@ -459,7 +459,13 @@ update_boot_offset (EmerPersistentCache *self,
         return save_timing_metadata (self, &relative_time, &absolute_time, NULL,
                                      NULL, NULL, error);
 
-      return TRUE;
+      /*
```
Kurt-von-Laven: nit: Prefer

```
/* GNOME comment
 * style
 */
```

for new comments.

cosimoc: The rest of the file uses a different style AFAICS so I will leave it like this unless you strongly object.

Kurt-von-Laven: WFM

daemon/emer-persistent-cache.c
```diff
@@ -459,7 +459,13 @@ update_boot_offset (EmerPersistentCache *self,
...
+       * Always make sure a file exists with the boot offset to maintain the
+       * invariant that when this function returns TRUE, a keyfile exists on
+       * disk.
+       */
+      return save_timing_metadata (self, NULL, NULL, &priv->boot_offset,
```
Kurt-von-Laven: We can't write to disk here because this code path gets hit every time an event is recorded. In particular, Roddy's constraint for the metrics system is that we write no more than once an hour. About to hop into a 1:1, but I can investigate why the test was failing after that if you like.

daemon/emer-persistent-cache.c
```diff
@@ -459,7 +459,13 @@ update_boot_offset (EmerPersistentCache *self,
         return save_timing_metadata (self, &relative_time, &absolute_time, NULL,
```
Kurt-von-Laven: nit: "missign" --> "missing" in the last commit message.

daemon/emer-persistent-cache.c
Kurt-von-Laven: Just realized why this is happening. I wasn't careful enough in my previous review. It's the test that is wrong, not the production code. We can drop commit 9a06e479ee6d822ba52d35d4c07d9e1be28c252f and modify test_persistent_cache_get_offset_can_build_boot_metadata_file so that it calls `make_testing_cache` and unrefs the persistent cache, then unlinks the boot offset metadata file, and then calls `make_testing_cache` again. I am pretty comfortable only handling the case where the boot offset metadata file mysteriously went missing at initialization especially since we already write the boot offset metadata file hourly and at shutdown. Arguably, we don't need to handle it at all, but that would really explode the scope of this PR.

Kurt-von-Laven:
Revise ready!

daemon/emer-persistent-cache.c
cosimoc: The reason the test fails is written in the commit message, isn't it?
On one hand the persistent cache code does not monitor the keyfile, and when the test unlinks it, this method will just return TRUE if the offset is already initialized.
On the other hand the test expects the file to be re-created every time `get_boot_offset()` is created. If we don't do this, we should remove that expectation from the test.

cosimoc: Oh, didn't see your second comment here. That matches my proposal too. Will do.

Kurt-von-Laven: Cool!

daemon/emer-persistent-cache.c
```diff
@@ -778,8 +778,11 @@ emer_persistent_cache_finalize (GObject *object)
       g_error_free (error);
```
Kurt-von-Laven: In the last commit message did you mean a "persistent" cache rather than a "permanent" cache?

tests/daemon/test-persistent-cache.c
```diff
@@ -1186,12 +1156,12 @@ main (gint                argc,
...
   ADD_CACHE_TEST_FUNC ("/persistent-cache/reads-cached-boot-offset",
                        test_persistent_cache_reads_cached_boot_offset);
-  ADD_CACHE_TEST_FUNC ("/persistent-cache/get-offset-wont-update-timestamps-if-it-isnt-supposed-to",
-                       test_persistent_cache_get_offset_wont_update_timestamps_if_it_isnt_supposed_to);
   ADD_CACHE_TEST_FUNC ("/persistent-cache/updates-timestamps-on-finalize",
                        test_persistent_cache_updates_timestamps_on_finalize);
   ADD_CACHE_TEST_FUNC ("/persistent-cache/get-offset-can-build-boot-metadata-file",
-                       test_persistent_cache_get_offset_can_build_boot_metadata_file);
+                       test_persistent_cache_can_build_boot_metadata_file);
+  ADD_CACHE_TEST_FUNC ("/persistent-cache/reads-cached-boot-offset-when-initialized",
```
Kurt-von-Laven: nit: If we keep this test, it would be nice to list it in the same order that it appears above (right below `test_persistent_cache_reads_cached_boot_offset`).

tests/daemon/test-persistent-cache.c
```diff
@@ -1041,33 +1018,37 @@ test_persistent_cache_reads_cached_boot_offset (gboolean     *unused,
...
 /*
- * Ensures that a request for the boot time offset doesn't cause a write to the
- * metadata file if the 'always_update_timestamps' parameter is set to FALSE.
+ * Ensures that after initialization the persistent cache will not read the
+ * metadata file again when calling get_boot_time_offset().
  */
 static void
-test_persistent_cache_get_offset_wont_update_timestamps_if_it_isnt_supposed_to (gboolean     *unused,
-                                                                                gconstpointer dontuseme)
+test_persistent_cache_reads_cached_boot_offset_when_initialized (gboolean *unused,
```
Kurt-von-Laven: As far as I can tell, the only difference between this test and `test_persistent_cache_reads_cached_boot_offset` is that this test unlinks the file on disk rather than mangling it. I feel like we probably don't need both tests, but if we do keep both, these tests should be named in a way that indicates their relationship to one another.

tests/daemon/test-persistent-cache.c
```diff
@@ -1041,33 +1018,37 @@ test_persistent_cache_reads_cached_boot_offset (gboolean     *unused,
...
   GError *error = NULL;
-
-  // This call shouldn't update the metadata file.
   gboolean get_succeeded =
-    emer_persistent_cache_get_boot_time_offset (cache, NULL, &error);
+    emer_persistent_cache_get_boot_time_offset (cache, &first_offset, &error);
+  g_assert_no_error (error);
+  g_assert_true (get_succeeded);
+
+  // Remove the boot offset file
```
Kurt-von-Laven: nit: Terminal period.

tests/daemon/test-persistent-cache.c
```diff
@@ -1106,26 +1080,22 @@ test_persistent_cache_updates_timestamps_on_finalize (gboolean     *unused,
...
+                                                    gconstpointer dontuseme)
 {
   EmerPersistentCache *cache = make_testing_cache ();
+  g_object_unref (cache);
 
-  /*
-   * Don't write a default boot offset file. We want to create a new one via
-   * production code.
-   */
+  // Remove the boot offset file
```
Kurt-von-Laven: nit: Terminal period.

Kurt-von-Laven:
Revise ready!

daemon/emer-persistent-cache.c
cosimoc: Indeed - good catch!

tests/daemon/test-persistent-cache.c
```diff
@@ -1041,33 +1018,39 @@ test_persistent_cache_reads_cached_boot_offset (gboolean     *unused,
...
- * metadata file if the 'always_update_timestamps' parameter is set to FALSE.
+ * Tests that the in-memory cache is used when present by removing the boot
+ * offset from disk between calls to the same instance of a persistent cache.
+ * This test exercises a code path similar to
+ * test_persistent_cache_reads_cached_boot_offset_when_file_changed().
  */
 static void
-test_persistent_cache_get_offset_wont_update_timestamps_if_it_isnt_supposed_to (gboolean     *unused,
-                                                                                gconstpointer dontuseme)
+test_persistent_cache_reads_cached_boot_offset_when_file_missing (gboolean *unused,
```
Kurt-von-Laven: nit: Vertical alignment.

tests/daemon/test-persistent-cache.c
```diff
@@ -1041,33 +1018,39 @@ test_persistent_cache_reads_cached_boot_offset (gboolean     *unused,
...
+  g_assert_no_error (error);
+  g_assert_true (get_succeeded);
+
+  // Remove the boot offset file.
+  g_assert_cmpint (g_unlink (TEST_DIRECTORY BOOT_OFFSET_METADATA_FILE), ==, 0);
 
+  gint64 second_offset;
+
+  /*
+   * This call should read the offset from its in-memory cache, not the new one
```
Kurt-von-Laven: Maybe just remove ", not the new one on disk" since that only applies in the case of the other test.

Kurt-von-Laven:
Revise ready!

tests/daemon/test-persistent-cache.c
```diff
@@ -1084,13 +1064,6 @@ test_persistent_cache_updates_timestamps_on_finalize (gboolean     *unused,
 
   GError *error = NULL;
```
Kurt-von-Laven: nit: Variable is no longer used.

cosimoc: Fixed

Kurt-von-Laven:
**CR Pass**

PR #208, opened 2015-11-18
Fix crash at shutdown.

Kurt-von-Laven:
The metrics instrumentation daemon expects to be able to send messages
to the event recorder daemon as long as the event recorder daemon is
alive. The event recorder daemon is a D-Bus service, so it can only
receive messages while D-Bus is around to deliver them. Therefore,
instruct systemd to start D-Bus before and shut it down after the event
recorder daemon.

[endlessm/eos-sdk#3848]

### DIFF
--- a/daemon/emer-main.c
+++ b/daemon/emer-main.c
@@ -209,6 +209,7 @@ on_bus_acquired (GDBusConnection *system_bus,
                  EmerDaemon      *daemon)
 {
   EmerEventRecorderServer *server = emer_event_recorder_server_skeleton_new ();
+
   g_signal_connect (server, "handle-record-singular-event",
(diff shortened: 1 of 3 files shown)


Kurt-von-Laven:
Only the first commit is intended to be cherry-picked to eos2.5.

Kurt-von-Laven:
@cosimoc should we include this in 2.5.0? I wonder if it has been causing us to drop some percentage of shutdown events.

cosimoc:
@Kurt-von-Laven 2.5.0 is done at this point, but we will cherry-pick this for 2.5.1.

Kurt-von-Laven:
Roger.

PR #214, opened 2016-12-07
Gracefully handle invalid/corrupt data in the persistent cache

mariospr:
This patch set improves the way we deal with situations where invalid/corrupt data makes it to the persistent cache file in three ways:
  * Discard invalid data found after a valid chunk when reading from the cache, and just report that.
  * Purge the whole persistent cache file in cases where the corruption can't be detected earlier.
  * Send metrics events to the server so that we can track down when these situations happen

Note that is still unclear to me how we actually ended up in this scenario, which is still the big question here, but these changes should at least prevent the daemon from crashing and make it able to keep recording and sending metrics after having detected and deal with these kind of situations.

https://phabricator.endlessm.com/T12739

### DIFF
--- a/daemon/emer-circular-file.c
+++ b/daemon/emer-circular-file.c
@@ -698,6 +698,7 @@ emer_circular_file_read (EmerCircularFile *self,
                          gsize             data_bytes_to_read,
                          gsize            *num_elems,
                          guint64          *token,
+                         gboolean         *has_invalid,
                          GError          **error)
 {
   EmerCircularFilePrivate *priv =
@@ -728,6 +729,8 @@ emer_circular_file_read (EmerCircularFile *self,
   guint64 curr_data_bytes = 0;
   guint64 curr_disk_bytes = 0;
   GInputStream *input_stream = G_INPUT_STREAM (file_input_stream);
+
+  *has_invalid = FALSE;
   while (curr_disk_bytes < priv->size)
     {
       guint64 elem_size;
@@ -744,6 +747,7 @@ emer_circular_file_read (EmerCircularFile *self,
           g_warning ("Discarding invalid data found after byte %" G_GINT64_FORMAT,
                      (priv->head + curr_disk_bytes) % priv->max_size);
           set_metadata (self, curr_disk_bytes, priv->head, error);
+          *has_invalid = TRUE;
           break;
         }
 
--- a/daemon/emer-circular-file.c
+++ b/daemon/emer-circular-file.c
@@ -734,6 +734,19 @@ emer_circular_file_read (EmerCircularFile *self,
       if (!read_elem_size (self, input_stream, &elem_size, error))
         goto handle_failed_read;
 
+      /* Reading a zero-sized element here means that we have invalid
+       * data ahead, in which case we need to update the priv->size
+       * pointer so that the next time this is run it does not include
+       * the region of invalid data existing after this point.
+       */
+      if (elem_size == 0)
+        {
+          g_warning ("Discarding invalid data found after byte %" G_GINT64_FORMAT,
+                     (priv->head + curr_disk_bytes) % priv->max_size);
+          set_metadata (self, curr_disk_bytes, priv->head, error);
+          break;
+        }
+
       guint64 next_data_bytes = curr_data_bytes + elem_size;
       if (next_data_bytes > data_bytes_to_read)
         break;
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -574,7 +574,22 @@ add_stored_events_to_builders (EmerDaemon        *self,
                                 &num_variants, token, &error);
   if (!read_succeeded)
     {
-      g_warning ("Could not read from persistent cache: %s.", error->message);
+      if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA))
+        {
+          GError *local_error = NULL;
+
+          g_critical ("Invalid data read from the persistent cache. Clearing everything...");
+          if (!emer_persistent_cache_remove_all (priv->persistent_cache, &local_error))
+            {
+              g_warning ("Error removing data from the persistent cache: %s", local_error->message);
+              g_error_free (local_error);
+            }
+        }
+      else
+        {
+          g_warning ("Could not read from persistent cache: %s.", error->message);
+        }
+
       g_error_free (error);
       *read_bytes = 0;
       *token = 0;
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -108,6 +108,17 @@
   "uploading is disabled. You may enable uploading by setting " \
   "uploading_enabled to true in " PERMISSIONS_FILE
 
+/* Event ID to send a metric event when the cache has been found to be corrupt
+ * resulting in the removal of all its data to get it back to a clear state.
+ * This event will not include any useful payload (just an empty one). */
+#define CACHE_IS_CORRUPT_EVENT_ID "d84b9a19-9353-73eb-70bf-f91a584abcbd"
+
+/* Event ID to send a metric event when some elements in the cache are invalid.
+ * The payload for this event will be formated as a '(tt)' GVariant containing
+ * the number of valid elements found and the number of bytes read.
+ */
+#define CACHE_HAS_INVALID_ELEMENTS_EVENT_ID "cbfbcbdb-6af2-f1db-9e11-6cc25846e296"
+
 typedef struct _NetworkCallbackData
 {
   EmerDaemon *daemon;
@@ -117,6 +128,13 @@ typedef struct _NetworkCallbackData
   gint attempt_num;
   GTask *upload_task;
 } NetworkCallbackData;
+
+typedef struct _CacheMetricEventData
+{
+  EmerDaemon *daemon;
+  const char *event_id;
+  GVariant *payload;
+} CacheMetricEventData;
 
 typedef struct _EmerDaemonPrivate
 {
@@ -138,6 +156,7 @@ typedef struct _EmerDaemonPrivate
   gchar *server_uri;
 
   guint upload_events_timeout_source_id;
+  guint report_invalid_cache_data_source_id;
 
   EmerMachineIdProvider *machine_id_provider;
   EmerNetworkSendProvider *network_send_provider;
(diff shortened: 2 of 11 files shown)


daemon/emer-circular-file.c
```diff
@@ -734,6 +734,19 @@ emer_circular_file_read (EmerCircularFile *self,
       if (!read_elem_size (self, input_stream, &elem_size, error))
         goto handle_failed_read;
 
+      /* Reading a zero-sized element here means that we have invalid
+       * data ahead, in which case we need to update the priv->size
+       * pointer so that the next time this is run it does not include
+       * the region of invalid data existing after this poing.
```
ptomato: typo: poing -> point

daemon/emer-daemon.c
```diff
@@ -574,7 +574,22 @@ add_stored_events_to_builders (EmerDaemon        *self,
                                 &num_variants, token, &error);
   if (!read_succeeded)
     {
-      g_warning ("Could not read from persistent cache: %s.", error->message);
+      if (error->domain == G_IO_ERROR && error->code == G_IO_ERROR_INVALID_DATA)
```
cosimoc: Use `g_error_matches()`

daemon/emer-persistent-cache.c
```diff
@@ -1051,29 +1051,47 @@ emer_persistent_cache_read (EmerPersistentCache *self,
...
+  read_succeeded =
     emer_circular_file_read (priv->variant_file, &elems, cost, &num_elems,
                              &local_token, error);
   if (!read_succeeded)
     return FALSE;
 
-  GVariant **local_variants = g_new (GVariant *, num_elems);
-  for (gsize i = 0; i < num_elems; i++)
+  corrupt_data = FALSE;
+  local_variants = g_new (GVariant *, num_elems);
```
cosimoc: Declare these two variables here (I know the coding style here is a bit different from the usual C, but we should keep it consistent)

daemon/emer-persistent-cache.c
```diff
@@ -1086,6 +1104,28 @@ emer_persistent_cache_read (EmerPersistentCache *self,
...
+      while (j < num_elems)
+        g_bytes_unref (elems[j++]);
+      g_free (elems);
+
+      while (j < i)
+        g_variant_unref (local_variants[j++]);
+
+      *variants = NULL;
+      *num_variants = 0;
+      *token = 0;
```
cosimoc: There's no reason to set these when returning `FALSE`

daemon/emer-persistent-cache.c
```diff
@@ -1086,6 +1104,28 @@ emer_persistent_cache_read (EmerPersistentCache *self,
...
     }
 
+  if (corrupt_data)
+    {
+      /* Somehow the circular file contains corrupt data, meaning that we
+       * can't trust its contents anymore so we clean up and return an error. */
+      gsize j = 0;
+
+      while (j < num_elems)
+        g_bytes_unref (elems[j++]);
```
cosimoc: The bytes in `elems` with index between `0` and `i - 1` have already been unreferenced in the for loop above before encountering the corrupt data. So I think that `j` should be initialized to the value of `i` before the cycle, not `0`.

daemon/emer-persistent-cache.c
```diff
@@ -1086,6 +1104,28 @@ emer_persistent_cache_read (EmerPersistentCache *self,
...
+      g_free (elems);
+
+      while (j < i)
+        g_variant_unref (local_variants[j++]);
```
cosimoc: I think `j` should be initialized to `0` before this loop, since the variants to free in `local_variants` have index between `0` and `i - 1`

daemon/emer-daemon.c
mariospr: Ok

daemon/emer-persistent-cache.c
mariospr: Oh! My mistake. I was under the impression that it was using the C89 way of declaring everything at the beginning of the block. will change it

daemon/emer-persistent-cache.c
mariospr: That was actually my intention, will fix

daemon/emer-persistent-cache.c
mariospr: another mistake, will fix

daemon/emer-persistent-cache.c
mariospr: You'r right, I'll remove it

mariospr:
@cosimoc I addressed your comments and this is now ready for re-review. In the meantime I'll work in the alternative version of it that also sends a metric event when this problem is found, but I'll keep this around in case it's important to merge something today and I can't have the other one ready yet.

daemon/emer-daemon.c
```diff
@@ -548,6 +566,89 @@ add_events_to_builders (GVariant       **events,
...
+{
+  CacheMetricEventData *callback_data = g_new (CacheMetricEventData, 1);
+  callback_data->daemon = self;
+  callback_data->event_id = event_id;
+  callback_data->payload = payload != NULL ? g_variant_ref (payload) : NULL;
+
+  /* Do the report in a new iteration of the main loop to make sure we don't
+   * report the event before having finished processing the current cache.
+   */
+  g_idle_add ((GSourceFunc) report_invalid_data_in_cache_on_idle, callback_data);
```
cosimoc: Should probably save the source ID and cancel the timeout in case the daemon is destroyed.

daemon/emer-daemon.c
```diff
@@ -596,6 +699,18 @@ add_stored_events_to_builders (EmerDaemon        *self,
       return TRUE;
     }
 
+  if (has_invalid)
+    {
+      GVariant *payload = g_variant_new ("(tt)", num_variants, token);
+      g_variant_ref_sink (payload);
```
cosimoc: I don't think this is needed... `emer_daemon_record_singular_event()` will consume the floating ref

daemon/emer-daemon.c
```diff
@@ -548,6 +566,89 @@ add_events_to_builders (GVariant       **events,
...
+
+static void
+report_invalid_data_in_cache (EmerDaemon *self,
+                              const char *event_id,
+                              GVariant   *payload)
+{
+  CacheMetricEventData *callback_data = g_new (CacheMetricEventData, 1);
+  callback_data->daemon = self;
+  callback_data->event_id = event_id;
+  callback_data->payload = payload != NULL ? g_variant_ref (payload) : NULL;
```
cosimoc: I don't think you need to ref this, unless you want to use the variant after the idle.

daemon/emer-daemon.c
```diff
@@ -548,6 +566,89 @@ add_events_to_builders (GVariant       **events,
...
+
+  if (payload != NULL)
+    {
+      actual_payload = payload;
+    }
+  else
+    {
+      GVariant *unboxed_variant = g_variant_new_boolean (FALSE);
+      GVariant *empty_payload = g_variant_new_variant (unboxed_variant);
+      actual_payload = g_variant_ref_sink (empty_payload);
```
cosimoc: I don't think this is needed.

daemon/emer-daemon.c
```diff
@@ -548,6 +566,89 @@ add_events_to_builders (GVariant       **events,
...
+
+  emer_daemon_record_singular_event (self,
+                                     getuid (),
+                                     event_id_variant,
+                                     relative_time,
+                                     payload != NULL,
+                                     actual_payload);
+ free:
+  if (actual_payload != NULL)
+    g_variant_unref (actual_payload);
```
cosimoc: There should be no need to unref this if you follow my other suggestions

daemon/emer-daemon.c
mariospr: Are you sure about that? I can see get_nullable_payload() consuming the floating ref only for the case where `has_payload == FALSE` (meaning that it sinks the ref of the "empty auxiliary payload" is it calls it), which won't be the case here as we will be passing a specific payload whose ref needs sinking at some point.

daemon/emer-daemon.c
mariospr: Sure, will do.

daemon/emer-daemon.c
mariospr: In general I like to ref objects when sending them as data for asynchronous callbacks, but it's true in this particular case the idiom is not strictly needed since no one will be unreffing

daemon/emer-daemon.c
mariospr: I'm not completely sure either, I borrowed this [idiom from emtr_event_recorder_init](https://github.com/endlessm/eos-metrics/blob/27e4194ed5ca1c5966261ef1fed5fa182a4285a8/eosmetrics/emtr-event-recorder.c#L172), as that's how it constructs the empty payload for some reason, and assumed that was the right thing to do as I can't see the server side of things. But I'm happy to write it simpler if possible

daemon/emer-daemon.c
mariospr: I'm not sure really. I think it's safer to call g_variant_ref_sink() only (i.e. no other calls to g_variant_ref in other places) for each of the two cases [in the if above this one](https://github.com/endlessm/eos-event-recorder-daemon/pull/214/commits/4a5c09295d8875669b3bd98d7f8025348f184228#diff-3caff38bd400df071bb8b3e5d6ed8d96R611) and then unref here.

daemon/emer-daemon.c
```diff
@@ -1115,6 +1248,7 @@ emer_daemon_finalize (GObject *object)
   EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
 
   g_source_remove (priv->upload_events_timeout_source_id);
+  g_source_remove (priv->report_invalid_cache_data_source_id);
```
cosimoc: You have to check if this is not zero before calling this I think.

daemon/emer-daemon.c
```diff
@@ -548,6 +567,92 @@ add_events_to_builders (GVariant       **events,
...
+                 "install uuid-runtime first.", unparsed_event_id);
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
+report_invalid_data_in_cache_on_idle (CacheMetricEventData *callback_data)
+{
```
cosimoc: Need to set the source ID to zero in this callback

daemon/emer-daemon.c
cosimoc: I'm pretty sure that `emer_daemon_record_singular_event()` consumes floating refs like e.g. gdbus methods do. See e.g. https://github.com/endlessm/eos-desktop/blob/master/src/shell-app-system.c#L530

mariospr: Ok. I'll believe you :)

daemon/emer-daemon.c
mariospr: ok

PR #215, opened 2017-04-20
Change default configuration to "collect, don't upload"

wjt:
With the previous defaults, metrics collected during first boot would be dropped, even if the user consented to metrics in the FBE. This is problematic because the boot-time metrics include crucial facts like the OS version are not submitted the first time an Endless OS installation is used. This also prevented us collecting metrics from live sessions, since the "is this a live session?" metric is also reported at boot time, and every boot is the first boot.

This PR changes the default from "don't collect, do upload" to "do collect, don't upload", and amends SetEnabled() to change both flags in unison.

https://phabricator.endlessm.com/T14065


wjt:
I think this branch actually hits the network while running its test suite…

PR #216, opened 2017-04-22
T14065: fix unit test races

ramcq:
This avoids a race in the tests by erasing the daemon's permissions file after each time it is read during the opt-in/opt-out test. This way, when it re-appears on disk, we know that the daemon has re-written it after a state change internally.

One change to the daemon itself is necessary to cause it to write it's internal fallback configuration to a file in the case that the permissions file is missing, otherwise it's impossible for the test to check that the daemon has the correct default state initially (the uploading flag is not visible over the D-Bus API).

https://phabricator.endlessm.com/T14065

### DIFF
--- a/tests/test-opt-out-integration.py
+++ b/tests/test-opt-out-integration.py
@@ -24,6 +24,7 @@ import os
 import shlex
 import subprocess
 import tempfile
+import time
 import unittest
 
 import dbusmock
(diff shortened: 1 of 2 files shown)


tests/test-opt-out-integration.py
```diff
@@ -148,11 +149,20 @@ def test_UploadEvents_fails_if_disabled(self):
             self.interface.UploadEvents()
 
     def _check_config_file(self, enabled, uploading_enabled):
+        # the config file is written asyncronously by the daemon,
```
rshuler: Nit: spelling of asynchronously

tests/test-opt-out-integration.py
```diff
@@ -148,11 +149,20 @@ def test_UploadEvents_fails_if_disabled(self):
...
+        # so may not exist at the start of a test
+        while not os.path.exists(self.config_file):
```
rshuler: Probably not a good idea to have a potentially unbounded while loop, as it will get stuck here forever if an error prevents the path from ever existing.

PR #218, opened 2017-04-25
daemon: discard pending events on SetEnabled(False)

wjt:
Plus some logging, error-handling and test improvements.

https://phabricator.endlessm.com/T16713

### DIFF
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -141,7 +141,11 @@ typedef struct _EmerDaemonPrivate
 {
   guint network_send_interval;
   GQueue *upload_queue;
-  gboolean uploading;
+
+  /* Non-NULL iff an upload is in flight. An owned reference to the
+   * cancellable held by NetworkCallbackData.upload_task.
+   */
+  GCancellable *current_upload_cancellable;
 
   SoupSession *http_session;
 
@@ -215,7 +219,8 @@ finish_network_callback (NetworkCallbackData *callback_data)
   g_object_unref (callback_data->upload_task);
   g_free (callback_data);
 
-  priv->uploading = FALSE;
+  g_clear_object (&priv->current_upload_cancellable);
+
   g_signal_emit (self, emer_daemon_signals[SIGNAL_UPLOAD_FINISHED], 0u);
 }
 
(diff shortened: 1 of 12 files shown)


tests/daemon/test-daemon.c
```diff
@@ -464,6 +468,20 @@ assert_uploading_disabled (Fixture *fixture)
 }
 
 static void
+assert_uploading_disabled (Fixture     *fixture)
+{
+  assert_upload_events_fails
+ (fixture, EMER_ERROR_UPLOADING_DISABLED);
```
ptomato: spurious newline?

daemon/emer-daemon.c
```diff
@@ -141,7 +142,8 @@ typedef struct _EmerDaemonPrivate
 {
   guint network_send_interval;
   GQueue *upload_queue;
-  gboolean uploading;
+  /* Non-NULL iff an upload is in flight */
+  GCancellable *current_upload_cancellable;
```
ptomato: Who owns this cancellable? Looks like you unref it when freeing the network callback data below

wjt: Every `GTask` in the queue has its own `GCancellable`. This field holds a new ref to the current task's cancellable:

```c
  priv->current_upload_cancellable = g_object_ref (
    g_task_get_cancellable (upload_task));
```

and, as you say, unrefs it when the task ends. But actually I think it should be cleared before returning a value to the application code. Right now the GTask returns, then the cancellable is cleared – but if the GTask's callback caused another task to be popped from the queue, the wrong cancellable would get cleared. In practice this doesn't happen – there's a connectivity check, and handle_network_monitor_can_reach is (presumably?) called back asynchronously, but I think I should make this situation impossible rather than doesn't-happen-by-accident.

tests/daemon/test-daemon.c
wjt: Oops!

daemon/emer-daemon.c
wjt: While looking at this I noticed that there's a theoretical case where `finish_network_callback` could call `g_signal_emit` on a freed `EmerDaemon`, so added a patch to fix that case to the top of the PR.

wjt:
I've addressed @ptomato's feedback (I added a better comment about the `GCancellable` lifecycle) and added one extra patch with a refcounting fix.

daemon/emer-daemon.c
```diff
@@ -988,7 +1016,7 @@ upload_events (EmerDaemon         *self,
   EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
 
   GTask *upload_task =
-    g_task_new (self, NULL /* GCancellable */, callback, user_data);
+    g_task_new (self, g_cancellable_new (), callback, user_data);
```
cosimoc: Are you not leaking a reference to the cancellable here? `g_task_new()` will take an extra ref to it.

wjt: Great catch, I'll prepare another PR.

PR #221, opened 2017-08-18
T16933

cosimoc:
https://phabricator.endlessm.com/T16933

### DIFF
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -22,6 +22,7 @@
 
 #include "emer-daemon.h"
 
+#include <math.h>
 #include <time.h>
 #include <uuid/uuid.h>
 
@@ -128,6 +129,7 @@ typedef struct _NetworkCallbackData
   gsize num_stored_events;
   gsize num_buffer_events;
   gint attempt_num;
+  guint backoff_timeout_source_id;
 } NetworkCallbackData;
 
 typedef struct _CacheMetricEventData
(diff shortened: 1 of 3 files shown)


daemon/emer-daemon.c
```diff
@@ -564,21 +593,12 @@ handle_http_response (SoupSession *http_session,
...
-          return;
-        }
-
-      g_variant_unref (callback_data->request_body);
-      callback_data->request_body = updated_request_body;
-      queue_http_request (upload_task);
+      guint random_backoff_interval =
+        get_random_backoff_interval (priv->rand, callback_data->attempt_num);
+      callback_data->backoff_timeout_source_id =
+        g_timeout_add (random_backoff_interval,
```
wjt: The advantage of using `g_timeout_add_seconds` here would be to (in principle) allow timeout coalescing across the system. Of course this would mean that the first timeout would always be 1 second, the second would be 2 or 3 with 50% prob each, the third would be 4, 5, 6, 7 evenly, etc. but given the initial push happens 1h after boot (which is essentially a random start time) I still doubt we'd get into a situation where all Endless OS machines are retrying in perfect sync.

I just checked [GLib's implementation](https://git.gnome.org/browse/glib/tree/glib/gmain.c#n4578) and the idea is to align timers to a pseudorandom µsec which is consistent within the machine/session. The offset is based on the hash of the session bus address, or hostname if there is no session. So for this daemon, it's actually shared across all Endless OS machines because this is a system daemon (no session bus) and the hostname is `endless`. Not ideal, really it should use the machine ID (if present, which it is on all machines with systemd and/or D-Bus) but, rather annoyingly, GDBusConnection knows how to read the machine ID but does not expose this over anything except `org.freedesktop.DBus.Peer.GetMachineId`. Anyway…

Up to you whether you think this is important.

cosimoc: Given that this only happens in error conditions that should not be common, I don't think the impact on power consumption (the main reason to use `g_timeout_add_seconds()` AFAICS) would be noticeable... But you're right, technically we don't need millisecond precision here, so I will change it.

PR #224, opened 2017-11-20
Recover from corrupt cache metadata or config file

wjt:
We observed a system where `/etc/metrics/cache-size.conf` and `/var/cache/metrics/variants.dat.metadata` were full of `NUL` bytes, to the length you'd expect if the file were properly written. When it encountered these files, the metrics daemon logged a warning for the first, then logged a fatal error and fell over for the second.

I believe this is due to surprising behaviour of `g_file_set_contents()` when the file to be overwritten does not exist: it does not `fsync()` the temporary file before `close()` and `rename()`. Is this a bug? Maybe. This application appears to be doing everything right (relatively speaking): it creates both files with `g_key_file_save_to_file()` which is meant to be the correct, safe thing to do.

Regardless of the cause, any systems in the wild which have hit this problem now have a broken metrics daemon, so we need to deal with the case where one or both of these keyfiles cannot be parsed. The sensible recovery options seem to me to be:

* `cache-size.conf`: since there's only one value in this file, just re-initialize it.
* `variants.dat.metadata`: without the metadata it's impossible to interpret the contents of `variants.dat`, so again we need to re-initialize the keyfile, which logically discards the contents of `variants.dat`. In this case, we want to log an event so we have some idea how often this happens in the wild.

`EmerCacheSizeProvider` was, to my eyes, really overcomplicated, so I'm afraid I couldn't stop myself replacing it with a single function.

https://phabricator.endlessm.com/T19953

### DIFF
--- a/daemon/emer-cache-size-provider.c
+++ b/daemon/emer-cache-size-provider.c
@@ -1,6 +1,6 @@
 /* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*- */
 
-/* Copyright 2015 Endless Mobile, Inc. */
+/* Copyright 2015-2017 Endless Mobile, Inc. */
 
 /*
  * This file is part of eos-event-recorder-daemon.
@@ -22,16 +22,6 @@
 
 #include "emer-cache-size-provider.h"
 
-typedef struct EmerCacheSizeProviderPrivate
-{
-  gchar *path;
-  guint64 max_cache_size;
-  gboolean data_cached;
-  GKeyFile *key_file;
-} EmerCacheSizeProviderPrivate;
-
-G_DEFINE_TYPE_WITH_PRIVATE (EmerCacheSizeProvider, emer_cache_size_provider, G_TYPE_OBJECT)
-
 /*
  * The filepath to the metadata file containing the maximum persistent cache
  * size.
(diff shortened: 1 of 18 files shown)


daemon/emer-cache-size-provider.c
```diff
@@ -44,199 +34,55 @@ G_DEFINE_TYPE_WITH_PRIVATE (EmerCacheSizeProvider, emer_cache_size_provider, G_T
...
+  if (error != NULL)
     {
+      /* If the file doesn't exist, silently create it. Otherwise, something
+       * was wrong with it:
+       */
       if (!g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
-        goto handle_failed_read;
-
+        {
+          g_warning ("Error reading cache size from %s: %s", path,
```
wjt: Looking again, I think we should ignore `G_KEY_FILE_ERROR_GROUP_NOT_FOUND` and `G_KEY_FILE_ERROR_KEY_NOT_FOUND` and just silently add the key.

daemon/emer-circular-file.c
```diff
@@ -490,19 +495,23 @@ emer_circular_file_initable_init (GInitable    *initable,
...
-    g_key_file_load_from_file (priv->metadata_key_file, metadata_filepath,
-                               G_KEY_FILE_NONE, &local_error);
-  g_free (metadata_filepath);
-  if (!load_succeeded)
+  g_autoptr(GError) local_error = NULL;
+
+  if (!priv->reinitialize &&
+      !g_key_file_load_from_file (priv->metadata_key_file,
+                                  priv->metadata_filepath, G_KEY_FILE_NONE,
+                                  &local_error))
```
wjt: Similarly here?

tests/daemon/test-circular-file.c
```diff
@@ -103,8 +103,10 @@ make_circular_file (Fixture *fixture,
                     guint64  max_size)
 {
   GError *error = NULL;
+  gboolean reinitialize = FALSE;
```
mariospr: It looks like you can't avoid this declaration and pass `FALSE` directly to `_new() `

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+  test_circular_file_broken_metadata_file (fixture, contents, length,
+                                           G_KEY_FILE_ERROR_KEY_NOT_FOUND);
+}
+
 gint
 main (gint                argc,
       const gchar * const argv[])
 {
   g_test_init (&argc, (gchar ***) &argv, NULL);
+  g_test_bug_base ("https://phabricator.endlessm.com/");
```
mariospr: nice

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+
+  /* Initialize the metadata file to the desired incorrect contents */
+  ret = g_file_set_contents (metadata_file_path, metadata_file_contents, metadata_file_length, &error);
+  g_assert_no_error (error);
+  g_assert_true (ret);
+
+  g_autoptr(EmerCircularFile) circular_file = NULL;
+
+  circular_file = emer_circular_file_new (fixture->data_file_path, max_size,
+                                          /* reinitialize */ FALSE,
```
mariospr: Nit: I normally see this inline comments added **to the right** of the value, not to the left, because of keeping everything aligned on the left, so maybe do it here too?. Mind you, it might be that I'm used to a different style, so feel free to ignore

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+                                          &error);
+  g_assert_null (circular_file);
+  g_assert_error (error, G_KEY_FILE_ERROR, expected_error_code);
+  g_clear_error (&error);
+
+  circular_file = emer_circular_file_new (fixture->data_file_path, max_size,
+                                          /* reinitialize */ TRUE,
```
mariospr: Same comment (nit) here

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+   */
+  const gchar * const NO_STRINGS[0];
+  guint64 token = assert_strings_read (circular_file, NO_STRINGS, 0);
+  g_assert_false (emer_circular_file_has_more (circular_file, token));
+
+  /* Adding new stuff to the file should work. */
+  const gchar * const STRINGS[] =
+    {
+      "Sneezy", "phylum", "Europe", "sloth", "guacamole", "data link",
+      "Colossus of Rhodes"
```
mariospr: May I suggest "Mint cake" too?

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+                                          /* reinitialize */ TRUE,
+                                          &error);
+  g_assert_no_error (error);
+  g_assert_nonnull (circular_file);
+
+  /* Verify that the file is logically empty, even though it's physically full
+   * of asterisks.
+   */
+  const gchar * const NO_STRINGS[0];
+  guint64 token = assert_strings_read (circular_file, NO_STRINGS, 0);
```
mariospr: I know this is not your code, and also it could be due to me being used to other naming conventions, but I found it surprisingly confusing that `assert_strings_read()` does actually more stuff than simply asserting the desired outcome/state. It would be more clear, IMHO, if it was named something like `read_strings_and_check()` or similar, so that it's clear that it performs both the action that needs to be tested + the actual test.

Not for this patch though, but maybe to consider for a separate cleanup commit?

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+  g_assert_false (emer_circular_file_has_more (circular_file, token));
+
+  /* Adding new stuff to the file should work. */
+  const gchar * const STRINGS[] =
+    {
+      "Sneezy", "phylum", "Europe", "sloth", "guacamole", "data link",
+      "Colossus of Rhodes"
+    };
+  gsize NUM_STRINGS = G_N_ELEMENTS (STRINGS);
+  assert_strings_saved (circular_file, STRINGS, NUM_STRINGS);
```
mariospr: Same comment than for `assert_strings_read()` here (consider renaming in a separate cleanup commit?)

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+
+static void
+test_circular_file_metadata_file_empty (Fixture      *fixture,
+                                        gconstpointer unused)
+{
+  /* If the metadata file exists but is empty, we should initialize it and
+   * consider the circular file itself to be empty.
+   */
+  test_circular_file_broken_metadata_file (fixture, "", 0,
+                                           G_KEY_FILE_ERROR_GROUP_NOT_FOUND);
```
mariospr: You'll have to update the expectation here if you finally silently deal with empty files and add the missing group/keys there, I think.

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+   * when the target file does not yet exist; poorly-documented semantics
+   * ensue.
+   *
+   * Regardless of cause, a reasonable recovery path is to re-initialize the
+   * metadata, assuming the file is empty.
+   */
+  g_test_bug ("T19953");
+
+  test_circular_file_broken_metadata_file (fixture, contents, length,
+                                           G_KEY_FILE_ERROR_GROUP_NOT_FOUND);
```
mariospr: Same here

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+{
+  const gchar contents[] = "[metadata]\nsize=1024\nhead=0\n";
+  const gssize length = sizeof contents;
+
+  /* max_size is missing; just assume the file is empty, since this will never
+   * happen in practice without other keys also being missing unless someone
+   * edits the file by hand and then you get to keep both pieces.
+   */
+  test_circular_file_broken_metadata_file (fixture, contents, length,
+                                           G_KEY_FILE_ERROR_KEY_NOT_FOUND);
```
mariospr: And here too

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+                                               gconstpointer unused)
+{
+  const gchar contents[] = "[metadata]\nmax_size=1024\nhead=27\n";
+  const gssize length = sizeof contents;
+
+  /* size is missing; we should recover by treating the file as empty because
+   * we do not know at what byte we should loop around.
+   */
+  test_circular_file_broken_metadata_file (fixture, contents, length,
+                                           G_KEY_FILE_ERROR_KEY_NOT_FOUND);
```
mariospr: Likewise

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+  ret = g_file_set_contents (metadata_file_path, metadata_file_contents, metadata_file_length, &error);
+  g_assert_no_error (error);
+  g_assert_true (ret);
+
+  g_autoptr(EmerCircularFile) circular_file = NULL;
+
+  circular_file = emer_circular_file_new (fixture->data_file_path, max_size,
+                                          /* reinitialize */ FALSE,
+                                          &error);
+  g_assert_null (circular_file);
```
mariospr: If you silently deal with missing groups/keys you'll receive a valid circular file here, so these assertions won't longer be valid, I think. Might need to refactor this code a bit in that case

tests/daemon/test-circular-file.c
```diff
@@ -609,11 +611,158 @@ test_circular_file_shrink (Fixture      *fixture,
...
+                                               gconstpointer unused)
+{
+  const gchar contents[] = "[metadata]\nmax_size=1024\nsize=1024\n";
+  const gssize length = sizeof contents;
+
+  /* size is missing; we should recover by treating the file as empty, because
+   * we don't know where the head was supposed to be.
+   */
+  test_circular_file_broken_metadata_file (fixture, contents, length,
+                                           G_KEY_FILE_ERROR_KEY_NOT_FOUND);
```
mariospr: And here as well. Actually, it looks like you might need to refactor these tests a bit to accomodate the different way of dealing with missing groups/keys, because the `expected_error` parameter seems like not very useful except for the "junk data" case :-)

daemon/emer-cache-size-provider.c
```diff
@@ -44,199 +34,61 @@ G_DEFINE_TYPE_WITH_PRIVATE (EmerCacheSizeProvider, emer_cache_size_provider, G_T
...
 
-  GError *error = NULL;
-  gboolean load_succeeded =
-    g_key_file_load_from_file (priv->key_file, priv->path, G_KEY_FILE_NONE,
-                               &error);
-  if (!load_succeeded)
+  if (!g_key_file_load_from_file (key_file, path, G_KEY_FILE_NONE,
+                                  &error))
     {
       if (!g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
```
mariospr: Following a similar pattern in other places (based on your own review), you might want to also exclude here cases where the file exists but the either the GROUP or the KEY can't be found, in which case you'd add default values?

tests/daemon/test-circular-file.c
wjt: Fair enough: the style you prefer is also the style used elsewhere in this file so I will adjust my style.

tests/daemon/test-circular-file.c
wjt: Yeah, it is nice. If you call `g_test_bug("T19953")` in a test and run the test suite with `--verbose` it shows up in the output:

```
GTest: run: /cache-size-provider/recovers-if-corrupt/nul
(MSG: Bug Reference: https://phabricator.endlessm.com/T19953)
GTest: result: OK
```

Whether it's actually more useful than a comment is another question but it's there so let's use it!

daemon/emer-cache-size-provider.c
wjt: Yes, I'll make that change.

tests/daemon/test-circular-file.c
wjt: A very good idea

tests/daemon/test-circular-file.c
wjt: Good point – I'll split this test helper in two.

tests/daemon/test-circular-file.c
wjt: Yes, great idea.

tests/daemon/test-circular-file.c
wjt: Right – I decided that this case and the ones below, when the file's actively malformed (rather than being empty or all `NUL` which is "morally empty"), are serious enough that we do still want to return an error, and have the daemon report an event and explicitly reinitialize. So I split this up.

wjt:
I've made the necessary changes to handle the "empty keyfile" and "all-NUL keyfile" cases silently, while still warning/returning an error in the "actively malformed" cases. I also think I've addressed your feedback – thanks!

tests/daemon/test-circular-file.c
mariospr: Sounds good to me

tests/daemon/test-circular-file.c
```diff
@@ -595,25 +597,249 @@ test_circular_file_shrink (Fixture      *fixture,
...
+{
+  /* Verify that the file is logically empty, even though it's physically full
+   * of asterisks.
+   */
+  const gchar * const NO_STRINGS[0];
+  guint64 token = read_strings_and_check (circular_file, NO_STRINGS, 0);
+  g_assert_false (emer_circular_file_has_more (circular_file, token));
+
+  /* Adding new entries to the file and reading them back should work. */
+  const gchar * const STRINGS[] = { "Kendal Mint Cake" };
```
mariospr: EXCELLENT! /cc @pwithnall

mariospr:
Thanks @wjt, I looked through the differences (with special emphasis on the updated test cases) and it all looks good to me now. Merging...

PR #225, opened 2017-11-29
Fix build/distcheck; enable more warnings more often

wjt:
In #224 I forgot to commit one file, and introduced three test failures when running `make distcheck`. Mea culpa. While fixing these, I fixed an existing error when running `make distcheck` as non-root, enabled many more compiler warnings, and fixed them too.

https://phabricator.endlessm.com/T19953

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -36,7 +36,7 @@ AM_CFLAGS = @STRICT_CFLAGS@
 # directory when doing 'make distcheck'.
 AM_DISTCHECK_CONFIGURE_FLAGS = \
 	--enable-strict-flags \
-	--with-systemdsystemunitdir=$(libdir)/systemd \
+	--with-systemdsystemunitdir='$${prefix}/lib/systemd/system' \
 	$(NULL)
 if EOS_ENABLE_COVERAGE
 AM_DISTCHECK_CONFIGURE_FLAGS += --enable-coverage --with-coverage-dir=@EOS_COVERAGE_DIR@
--- a/Makefile.am
+++ b/Makefile.am
@@ -30,12 +30,13 @@ ACLOCAL_AMFLAGS = -I m4
 EXTRA_DIST = README.md
 
 # Compiler flags
-AM_CFLAGS = @STRICT_CFLAGS@
+AM_CFLAGS = @WARN_CFLAGS@
+AM_LDFLAGS = @WARN_LDFLAGS@
 
 # Enable strict compiler flags and install systemd unit files to the specified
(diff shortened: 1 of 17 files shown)


tests/daemon/mock-cache-size-provider.c
```diff
@@ -0,0 +1,40 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*- */
+
+/* Copyright 2017 Endless Mobile, Inc. */
```
smspillaz: Probably needs an actual licence header, though if this is the style in the rest of the code, maybe not worth worrying about.

smspillaz: Oh, derp, its literally 3 lines below. Too hasty to comment!

tests/daemon/mock-persistent-cache.c
```diff
@@ -88,6 +88,7 @@ emer_persistent_cache_new (const gchar *directory,
   return self;
 }
 
+G_GNUC_NORETURN
 EmerPersistentCache *
 emer_persistent_cache_new_full (const gchar              *directory,
                                 guint64                   cache_size,
```
smspillaz: Not sure if this is a good candidate for `G_GNUC_NORETURN`, `g_assert_not_reached` is supposed to be inert when compiled in release mode.

Maybe just `return NULL;` afterwards?

smspillaz:
Just one comment.

tests/daemon/mock-persistent-cache.c
wjt: Great point. I'll replace this with `g_return_val_if_reached (NULL);` which always returns the value, even in `G_DISABLE_CHECKS` mode.

Makefile.am
```diff
@@ -30,13 +30,14 @@ ACLOCAL_AMFLAGS = -I m4
...
-AM_CFLAGS = @STRICT_CFLAGS@
+AM_CFLAGS = @WARN_CFLAGS@
+AM_LDFLAGS = @WARN_LDFLAGS@
 
 # Enable strict compiler flags and install systemd unit files to the specified
 # directory when doing 'make distcheck'.
 AM_DISTCHECK_CONFIGURE_FLAGS = \
-	--enable-strict-flags \
-	--with-systemdsystemunitdir=$(libdir)/systemd \
+	--enable-compile-warnings=error \
```
wjt: I also missed  `--enable-strict-flags` here before.

PR #227, opened 2018-01-09
T18983: Followup fixes

smspillaz:
Some follow-up fixes for https://phabricator.endlessm.com/T18983

The override machine ID was not being written out correctly, which means that it could not be parsed and loaded later by the metrics daemon. Also, we should clear out old metrics events associated with the old ID. We cannot guarantee that they will be uploaded, so just throw them away, for now.

### DIFF
--- a/daemon/emer-machine-id-provider.c
+++ b/daemon/emer-machine-id-provider.c
@@ -253,6 +253,22 @@ hyphenate_uuid (gchar *uuid_sans_hyphens)
                           uuid_sans_hyphens + 16, uuid_sans_hyphens + 20);
 }
 
+/*
+ * Returns a newly-allocated copy of uuid_with_hyphens with hyphens removed at
+ * the appropriate positions as defined by uuid_unparse(3).
+ * uuid_with_hyphens is expected to be exactly 36 bytes, excluding the terminal
+ * null byte.
+ * Any extra bytes are ignored.
+ * The returned string is guaranteed to be have a newline and be nul-terminated.
+ */
+static gchar *
+dehyphenate_uuid (gchar *uuid_with_hyphens)
+{
+  return g_strdup_printf ("%.8s%.4s%.4s%.4s%.12s\n", uuid_with_hyphens,
+                          uuid_with_hyphens + 9, uuid_with_hyphens + 14,
+                          uuid_with_hyphens + 19, uuid_with_hyphens + 24);
+}
+
 static gboolean
 read_one_machine_id (const gchar  *machine_id_path,
                      uuid_t        id,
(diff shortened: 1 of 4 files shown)


daemon/emer-machine-id-provider.c
```diff
@@ -414,7 +433,7 @@ write_tracking_id_file (const gchar  *path,
         }
     }
 
-  if (!g_file_set_contents (path, serialized_override_machine_id, -1, error))
+  if (!g_file_set_contents (path, dehyphenated_serialized_machine_id, -1, error))
```
wjt: Why didn't the test suite catch this bug? Can we add a test that would have failed before, and now passes?

tests/daemon/test-daemon.c
```diff
@@ -1484,14 +1484,7 @@ static void
...
                                            gconstpointer  unused)
 {
-  g_autoptr(GFileIOStream) stream = NULL;
   g_autoptr(GError) error = NULL;
-  g_autoptr(GFile) override_machine_id_file =
-    g_file_new_tmp ("machine-id-override-XXXXXX",
-                    &stream,
-                    &error);
-  g_autofree gchar *override_machine_id_file_path =
-    g_file_get_path (override_machine_id_file);
```
wjt: Why remove these? Shouldn't we be reading back the file and asserting it is exactly n hex characters long, with no hyphens?

smspillaz: The purpose of this test isn't to check the written out machine-id file. It is to check that a different machine ID came through in the metrics request to the original one. It'd be worth adding a new test to check the format of the machine-ID that got written though.

tests/daemon/test-machine-id-provider.c
```diff
@@ -153,6 +155,28 @@ test_machine_id_provider_can_get_id_override (MachineIdTestFixture *fixture,
...
+  emer_machine_id_provider_reset_tracking_id (id_provider, &error);
+  g_assert_no_error (error);
+
+  /* Read the override_machine_id_file_path using g_file_get_contents
+   * and check that its size matches what we would normally write to the
+   * file */
+  g_file_get_contents (fixture->override_machine_id_file_path, &contents, NULL, &error);
+  g_assert_no_error (error);
+
+  g_assert_cmpint (strlen (contents), ==, strlen (TESTING_OVERRIDE_ID));
```
wjt: Okay, that'll do it!

wjt:
hey bot, test this please

wjt:
The build failure is because I was compiling on a machine without a new-enough `eos-metrics`.

PR #232, opened 2019-04-30
tests: Fix a -Wsign-compare warning

pwithnall:
`num_bytes_read` has already been checked for negativity, so it’s safe
to cast it and treat it as unsigned in the later comparison.

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T26444

### DIFF
--- a/tests/daemon/test-daemon.c
+++ b/tests/daemon/test-daemon.c
@@ -243,7 +243,7 @@ append_bytes (GPollableInputStream *pollable_input_stream,
     g_pollable_input_stream_read_nonblocking (pollable_input_stream,
                                               destination, num_bytes_remaining,
                                               NULL /* GCancellable */, &error);
-  if (num_bytes_read == -1)
+  if (num_bytes_read < 0)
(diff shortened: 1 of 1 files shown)


tests/daemon/test-daemon.c
```diff
@@ -252,7 +252,7 @@ append_bytes (GPollableInputStream *pollable_input_stream,
 
   guint new_length = byte_array->len + num_bytes_read;
   g_byte_array_set_size (byte_array, new_length);
-  return num_bytes_read == num_bytes_remaining;
+  return (gsize) num_bytes_read == num_bytes_remaining;
```
wjt: I always have to look up the precedence of casts versus other operators. I sometimes write:

```suggestion
  return num_bytes_remaining == (gsize) num_bytes_read;
```

but I think this might fall under "if you have been writing C for ten years and still can't remember this simple fact, perhaps this is one of the few cases where C is not at fault" :p

PR #234, opened 2019-09-06
policy: Add action for eos-write-location

ptomato:
This allows eos-write-location to be run without prompting for admin
access. eos-write-location is a downstream script that is part of
gnome-initial-setup, which is used to save information about the Endless
Solutions deployment that a machine belongs to.

https://phabricator.endlessm.com/T27710

### DIFF
--- a/data/com.endlessm.Metrics.policy
+++ b/data/com.endlessm.Metrics.policy
@@ -48,4 +48,15 @@
     </defaults>
   </action>
 
+  <action id="com.endlessm.Metrics.WriteLocation">
+    <description>Write file describing Solutions deployment location</description>
+    <message>Authentication is required to set up metrics</message>
+    <defaults>
+      <allow_any>auth_admin</allow_any>
+      <allow_inactive>auth_admin</allow_inactive>
+      <allow_active>yes</allow_active>
+    </defaults>
+    <annotate key="org.freedesktop.policykit.exec.path">/usr/lib/gnome-initial-setup/eos-write-location</annotate>
+  </action>
+
 </policyconfig>


data/com.endlessm.Metrics.policy
```diff
@@ -48,4 +48,15 @@
     </defaults>
   </action>
 
+  <action id="com.endlessm.Metrics.WriteLocation">
+    <description>Write file describing Solutions deployment location</description>
+    <message>Authentication is required to set up metrics</message>
+    <defaults>
+      <allow_any>auth_admin</allow_any>
+      <allow_inactive>auth_admin</allow_inactive>
+      <allow_active>yes</allow_active>
```
wjt: “All active administrators can `mv` an arbitrary file to `/etc` and make it world-readable without authentication” seems quite a broad policy. (I eyeballed `/usr/lib/gnome-initial-setup/eos-write-location` and it doesn't quote the argument to `mv` so actually it's even broader: "pass arbitrary arguments to `mv` so long as the last one is `/etc/metrics/location.conf`".)

ptomato: I guess I wasn't thinking very clearly when I wrote this.

But this does bring me back to the question I raised in https://phabricator.endlessm.com/T27710#750975 where I don't understand why any user other than gnome-initial-setup needs to be able to do this.

I suppose it'd be better on the whole to make the file private to eos-event-recorder-daemon and add a DBus method to set the deployment ID instead of the eos-write-location script. I'll try to work on that after solving the freezing issue.

wjt: Are you certain that any user other than gnome-initial-setup **is** trying to run this?

ptomato: That was what the error message said in the first place: `Authenticating as test,,, (test)`, where `test` was Izzy's username

wjt: That doesn't mean that `test` is trying to run the command, only that it is `test` who is being asked to approve it. If you run `pkexec id` as an unprivileged user, eg `shared`, you'll get a polkit prompt for an (arbitrarily-picked?) non-system admin user. Since this call to `pkexec` happens after the user account is created, polkit picks the only non-system admin user, namely `test` which was just created.

ptomato: That clarifies things a lot, I would never have guessed that the answer to running pkexec as a non-admin user is not "no" but "well here, have a random admin user, maybe you know their password"!

PR #235, opened 2020-02-24
T28869

uajain:
This PR nukes the maintainence of two uuid (/etc/machine-id and a user-override tracking ID).
We only need one tracking ID which can be reset by the user.

1)  Adjust tests to use the new paradigm (tests pass locally :heavy_check_mark: )
2) Only create new ID if no tracking ID is found at TRACKING_ID_PATH (tested locally to notice a same uuid across reboot :heavy_check_mark: )

____

https://phabricator.endlessm.com/T28869

### DIFF
--- a/daemon/emer-machine-id-provider.c
+++ b/daemon/emer-machine-id-provider.c
@@ -54,9 +54,8 @@ G_DEFINE_TYPE_WITH_PRIVATE (EmerMachineIdProvider, emer_machine_id_provider, G_T
 
 /*
  * Filepath where an overridden random UUID, separate from /etc/machine-id
- * is stored. The machine-id might be read from this path and used as the
- * tracking ID in cases where we don't want to continue using the machine-id,
- * either on user request or when we enter the demo mode.
+ * is stored. This machine-id is read from this path and used as the
+ * default tracking ID for metrics purposes.
  */
 #define TRACKING_ID_PATH SYSCONFDIR "/metrics/tracking-id"
 
@@ -70,6 +69,8 @@ enum
 };
 
 static GParamSpec *emer_machine_id_provider_props[NPROPS] = { NULL, };
+
+static gboolean write_tracking_id_file (const gchar *path, GError **error);
 
 /*
  * SECTION:emer-machine-id-provider
@@ -304,10 +305,26 @@ read_machine_id (EmerMachineIdProvider *self)
 
   if (!read_one_machine_id (priv->tracking_id_path, id, &local_error))
     {
-      if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+      if (g_error_matches (local_error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
         {
-          g_debug ("Override machine id file %s does not exist.",
+          g_debug ("Tracking id file %s does not exist hence creating one.",
                    priv->tracking_id_path);
+          g_clear_error (&local_error);
+
+          if (!write_tracking_id_file (priv->tracking_id_path, &local_error))
+            {
+              g_message ("Failed to initialize tracking ID at %s: %s.",
+                         priv->tracking_id_path,
+                         local_error->message);
+              return FALSE;
+            }
+	  else if (!read_one_machine_id (priv->tracking_id_path, id, &local_error))
+            {
+              g_message ("Failed to read tracking id %s: %s",
+                         priv->tracking_id_path,
+                         local_error->message);
+              return FALSE;
+            }
         }
       else if (g_error_matches (local_error,
                                 EMER_ERROR,
@@ -316,9 +333,14 @@ read_machine_id (EmerMachineIdProvider *self)
           g_message ("Failed to read tracking id %s: %s",
                      priv->tracking_id_path,
                      local_error->message);
+          return FALSE;
         }
-
-      return FALSE;
+      else
+        {
+          g_message ("Error occured while reading tracking id at %s: %s",
+                     priv->tracking_id_path, local_error->message);
+          return FALSE;
+        }
     }
 
   uuid_copy (priv->id, id);
--- a/daemon/emer-machine-id-provider.c
+++ b/daemon/emer-machine-id-provider.c
@@ -36,8 +36,7 @@
 
 typedef struct EmerMachineIdProviderPrivate
 {
-  gchar *path;
-  gchar *override_path;
+  gchar *tracking_id_path;
   uuid_t id;
   gboolean id_is_valid;
 } EmerMachineIdProviderPrivate;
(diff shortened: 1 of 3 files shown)


daemon/emer-machine-id-provider.c
```diff
@@ -186,9 +187,18 @@ emer_machine_id_provider_class_init (EmerMachineIdProviderClass *klass)
 static void
 emer_machine_id_provider_init (EmerMachineIdProvider *self)
 {
+  g_autoptr(GError) local_error = NULL;
   EmerMachineIdProviderPrivate *priv =
     emer_machine_id_provider_get_instance_private (self);
 
+
+  if (!write_tracking_id_file (TRACKING_ID_OVERRIDE, &local_error))
```
wjt: This would cause the ID to be reset every time the system reboots. We don't want that – we want the ID to persist unless/until the user re-sets it. It should be written only if missing.

daemon/emer-machine-id-provider.c
```diff
@@ -186,9 +187,18 @@ emer_machine_id_provider_class_init (EmerMachineIdProviderClass *klass)
...
+    g_message ("Failed to initialize override machine id at %s: %s."
+               "Default machine id at %s will be used",
```
wjt: I also don't think we want this behaviour. If you look at `read_machine_id`, it returns `FALSE` if the machine ID can't be read.

I suggest removing all code that accesses `/etc/machine-id`.

uajain: > I suggest removing all code that accesses /etc/machine-id.

Ah yeah, I was going to ask you that. Great.

uajain:
@wjt I updated the PR, please have a look.

I know we discussed on slack my temptation to rename `emer-machine-id` to `emer-tracking-id`
but I think that should be done as a follow up PR to this to reconcile the naming convention again

daemon/emer-machine-id-provider.c
```diff
@@ -152,31 +133,51 @@ emer_machine_id_provider_finalize (GObject *object)
...
+    emer_machine_id_provider_get_instance_private (self);
+  g_autoptr(GError) local_error = NULL;
+  uuid_t uuid;
+
+  /* Check if tracking ID is already available. This make sure we do not re-create
+   * a new uuid everytime across reboots.
+   */
+  if (!emer_machine_id_provider_get_id (self, uuid))
+    {
+      if (!write_tracking_id_file (priv->tracking_id_path, &local_error))
```
wjt: Do the tests exercise this code path? I think not because `setup` always creates the file. The most important case is trying to read the ID //without// explicitly writing it, because that will be the initial state of all real systems. (Previously we have relied on `/etc/machine-id` existing.)

uajain: I added a new test in the suite which doesn't use `setup` so the tracking-id file is not explicitly written.

uajain:
retest this please

uajain:
retest this please

daemon/emer-machine-id-provider.c
```diff
@@ -333,35 +325,29 @@ read_machine_id (EmerMachineIdProvider *self)
...
 
-  if (!read_one_machine_id (priv->override_path, id, &local_error))
+  if (!read_one_machine_id (priv->tracking_id_path, id, &local_error))
     {
       if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
         {
-          g_debug ("Override machine id file %s does not exist, trying default.",
-                   priv->override_path);
+          g_debug ("Override machine id file %s does not exist.",
+                   priv->tracking_id_path);
```
wjt: Perhaps we should call `write_tracking_id_file` here, rather than in constructed?

daemon/emer-machine-id-provider.c
```diff
@@ -333,35 +325,29 @@ read_machine_id (EmerMachineIdProvider *self)
...
-      if (!read_one_machine_id (priv->path, id, &local_error))
-        {
-          g_message ("Failed to read machine id %s: %s",
-                     priv->path,
-                     local_error->message);
-        }
+      return FALSE;
     }
 
   if (uuid_is_null (id))
```
wjt: This error path is now unneeded because we will only reach this point if `read_one_machine_id` succeeds.

tests/daemon/test-machine-id-provider.c
```diff
@@ -125,73 +146,52 @@ test_machine_id_provider_new_succeeds (MachineIdTestFixture *fixture,
...
+test_machine_id_provider_can_get_tracking_id (MachineIdTestFixture *fixture,
                                               gconstpointer         dontuseme)
 {
   g_autoptr(EmerMachineIdProvider) id_provider =
-    emer_machine_id_provider_new_full (fixture->machine_id_file_path,
-                                       fixture->override_machine_id_file_path);
+    emer_machine_id_provider_new_full (fixture->tracking_id_file_path);
   uuid_t id;
-  write_testing_machine_id (fixture->override_machine_id_file_path, TESTING_OVERRIDE_ID);
+  write_testing_machine_id (fixture->tracking_id_file_path, TESTING_TRACKING_ID);
```
wjt: This is unnecessary because it is written in `setup`.

tests/daemon/test-machine-id-provider.c
```diff
@@ -125,73 +146,52 @@ test_machine_id_provider_new_succeeds (MachineIdTestFixture *fixture,
...
-  /* Read the override_machine_id_file_path using g_file_get_contents
+  /* Read the tracking_id_file_path using g_file_get_contents
    * and check that its size matches what we would normally write to the
    * file */
-  g_file_get_contents (fixture->override_machine_id_file_path, &contents, NULL, &error);
+  g_file_get_contents (fixture->tracking_id_file_path, &contents, NULL, &error);
   g_assert_no_error (error);
 
-  g_assert_cmpint (strlen (contents), ==, strlen (TESTING_OVERRIDE_ID));
+  g_assert_cmpint (strlen (contents), ==, strlen (TESTING_TRACKING_ID));
```
wjt: I know this is pre-existing code, but we should assert that the new ID is not the same as the old ID. (There is a 2 ** 128 chance of this failing but that's OK.)

We should also call `emer_machine_id_provider_get_id` and check that it returns a different ID.

tests/daemon/test-machine-id-provider.c
```diff
@@ -206,14 +206,16 @@ main (gint                argc,
...
-                       test_machine_id_provider_can_get_id_override_malformed);
+  ADD_CACHE_TEST_FUNC ("/machine-id-provider/can-get-tracking-id",
+                       test_machine_id_provider_can_get_tracking_id);
+  ADD_CACHE_TEST_FUNC ("/machine-id-provider/can-write-correctly-formed-tracking-id",
+                       test_machine_id_provider_writes_correctly_formed_tracking_id);
+  ADD_CACHE_TEST_FUNC ("/machine-id-provider/read-malformed-tracking-id",
+                       test_machine_id_provider_read_malformed_tracking_id);
+  /* Test when we do not write a tracking ID file explicitly while setup().
+   * emer_machine_id_provider_new_full should initialize a new tracking ID on its own. */
+  g_test_add ("/machine-id-provider/create-tracking-id-if-unavailable", MachineIdTestFixture, NULL,
```
wjt: I'd prefer for this to use `setup` just like the others – half of the test is boilerplate copied from `setup`. You could pass GINT_TO_POINTER (TRUE) as the `tdata` parameter (`NULL` here) to tell `setup` not to write the file.

dbnicholson:
bot, retest this please

daemon/emer-machine-id-provider.c
uajain: Good idea.

However, after this change, to detect that the tracking-id-file initially doesn't exist, I had to change 
`G_IO_ERROR_NOT_FOUND` to `G_FILE_ERROR_NO_ENT` in `read_machine_id()`  . I debugged this when the test supporting this change was failiing for no obvious reasons.

I will update the PR tomorrow.

tests/daemon/test-machine-id-provider.c
uajain: Addressed.

tests/daemon/test-machine-id-provider.c
uajain: addressed

daemon/emer-machine-id-provider.c
uajain: addressed

PR #236, opened 2020-02-25
daemon: Set tracking ID on DBus so that it can be read by client

uajain:
https://phabricator.endlessm.com/T28869

___

Depends on #235 

~~(Do not merge; compile test only)~~

### DIFF
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -1604,6 +1604,19 @@ emer_daemon_new (const gchar             *persistent_cache_directory,
                        NULL);
 }
 
+gchar *
+emer_daemon_get_tracking_id (EmerDaemon *self)
+{
+  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
+  g_autofree gchar *machine_id_hex = NULL;
+  uuid_t uuid;
+
+  if (emer_machine_id_provider_get_id (priv->machine_id_provider, &machine_id_hex, uuid))
+      return g_steal_pointer (&machine_id_hex);
+
+  return NULL;
+}
+
 gboolean
 emer_daemon_reset_tracking_id (EmerDaemon  *self,
                                GError     **error)
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -843,6 +843,7 @@ create_request_body (EmerDaemon *self,
 
   uuid_t machine_id;
   gboolean read_id = emer_machine_id_provider_get_id (priv->machine_id_provider,
+                                                      NULL,
                                                       machine_id);
   if (!read_id)
     {
--- a/daemon/emer-machine-id-provider.c
+++ b/daemon/emer-machine-id-provider.c
@@ -241,10 +241,11 @@ dehyphenate_uuid (gchar *uuid_with_hyphens)
 
 static gboolean
 read_one_machine_id (const gchar  *machine_id_path,
+                     gchar       **machine_id_hex,
                      uuid_t        id,
                      GError      **error)
 {
-  gchar *machine_id_sans_hyphens;
+  g_autofree gchar *machine_id_sans_hyphens = NULL;
   gsize machine_id_sans_hyphens_length;
   gboolean read_succeeded =
     g_file_get_contents (machine_id_path, &machine_id_sans_hyphens,
@@ -277,7 +278,6 @@ read_one_machine_id (const gchar  *machine_id_path,
     }
 
   gchar *hyphenated_machine_id = hyphenate_uuid (machine_id_sans_hyphens);
-  g_free (machine_id_sans_hyphens);
 
   gint parse_failed = uuid_parse (hyphenated_machine_id, id);
   g_free (hyphenated_machine_id);
@@ -292,18 +292,21 @@ read_one_machine_id (const gchar  *machine_id_path,
       return FALSE;
     }
 
+  if (machine_id_hex)
+    *machine_id_hex = g_strchomp (g_steal_pointer (&machine_id_sans_hyphens));
+
   return TRUE;
 }
 
 static gboolean
-read_machine_id (EmerMachineIdProvider *self)
+read_machine_id (EmerMachineIdProvider *self, gchar **machine_id_hex)
 {
   EmerMachineIdProviderPrivate *priv =
     emer_machine_id_provider_get_instance_private (self);
   g_autoptr(GError) local_error = NULL;
   uuid_t id;
 
-  if (!read_one_machine_id (priv->tracking_id_path, id, &local_error))
+  if (!read_one_machine_id (priv->tracking_id_path, machine_id_hex, id, &local_error))
     {
       if (g_error_matches (local_error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
         {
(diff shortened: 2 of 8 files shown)


uajain:
@wjt mind eye-balling this too (apart from #235)?

daemon/emer-daemon.c
```diff
@@ -1255,6 +1255,14 @@ set_max_bytes_buffered (EmerDaemon *self,
...
 }
 
+static gchar *
+unhyphenate_uuid (gchar *uuid_with_hyphens)
+{
+  return g_strdup_printf ("%.8s%.4s%.4s%.4s%.12s\n", uuid_with_hyphens,
+                          uuid_with_hyphens + 9, uuid_with_hyphens + 14,
+                          uuid_with_hyphens + 19, uuid_with_hyphens + 24);
+}
+
```
wjt: I recognise this code from the tests. Perhaps it could live in metrics-shared.[ch]?

daemon/emer-main.c
```diff
@@ -108,13 +108,18 @@ on_reset_tracking_id (EmerEventRecorderServer *server,
                       EmerDaemon              *daemon)
 {
   g_autoptr(GError) error = NULL;
+  const gchar *tracking_id;
```
wjt: Leaks

uajain:
> 
> Hyphenating and parsing the ID as a UUID does serve to validate it. Could we do something like, change the signature of the accessor function to return it in both parsed and unparsed form:
> 
> ```
> gboolean               emer_machine_id_provider_get_id   (EmerMachineIdProvider *self,
>                                                           char                   machine_id_hex[33]);
>                                                           uuid_t                 machine_id);
> ``` 
> 
> and have the implementation copy the unparsed ID into the first out parameter, once it has been validated, and set machine_id_hex[32] to `'\0'`. Or return an allocated string rather than have the caller pass in an array.

I will try to address this as a part of this PR. Keeping #235's scope limited as it is.

daemon/emer-machine-id-provider.c
```diff
@@ -292,18 +292,21 @@ read_one_machine_id (const gchar  *machine_id_path,
       return FALSE;
     }
 
+  if (machine_id_hex)
+    g_strlcpy (machine_id_hex, machine_id_sans_hyphens, sizeof(gchar)*FILE_LENGTH);
```
wjt: In full knowledge that I originally suggested this array parameter approach, I think it would be better to accept a `char **` parameter and here use:

```suggestion
    *machine_id_hex = g_strchomp (g_steal_pointer (&machine_id_sans_hyphens));
```

on the basis that we have already allocated that string, and it saves having to think about whether the inbound array is big enough.

(For future reference, `sizeof(char) == 1` is always true because `sizeof` is defined to return the number of `char`s needed to store the given expression or type.)

uajain: > In full knowledge that I originally suggested this array parameter approach, I think it would be better to accept a char ** parameter

So should I dynamically allocate the array right from the start? I think array allocated on function's stack(as of now) will not be able able passed upon without forced `(char **)`  pointer-conversion to the subsequent call functions

wjt: Give the innermost function this signature:

```c
static gboolean
read_one_machine_id (const gchar  *machine_id_path,
                     gchar       **machine_id_hex,
                     uuid_t        id,
                     GError      **error)
```

And make the same change to all intervening function's signatures.

Change this line to:

```suggestion
    *machine_id_hex = g_strchomp (g_steal_pointer (&machine_id_sans_hyphens));
```

In `emer_daemon_get_tracking_id` define:

```c
   gchar *machine_id_hex = NULL;
```

Call:

```c
   if (emer_machine_id_provider_get_id (priv->machine_id_provider, &machine_id_hex, uuid)) ```

and then return it to the caller.

Comment (uajain) [reply daemon/emer-machine-id-provider.c]: addressed

File: daemon/emer-daemon.c [hunk 4]
```diff
@@ -1603,6 +1604,22 @@ emer_daemon_new (const gchar             *persistent_cache_directory,
...
+gchar *
+emer_daemon_get_tracking_id (EmerDaemon *self)
+{
+  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
+  g_autofree gchar *machine_id_hex = NULL;
+  uuid_t uuid;
+
+  if (emer_machine_id_provider_get_id (priv->machine_id_provider, &machine_id_hex, uuid))
+    {
+      machine_id_hex[32] = '\0';
```
wjt: Can we move stripping the newline into `emer_machine_id_provider_get_id`?

wjt: Oh, we did, so this line should not be needed.

tests/daemon/test-machine-id-provider.c
```diff
@@ -153,11 +153,20 @@ test_machine_id_provider_can_get_tracking_id (MachineIdTestFixture *fixture,
...
+  // simultaneous check if we can get a unparsed_id directly to show in UI
+  g_autofree gchar *unparsed_id_direct = NULL;
+
+  g_assert (emer_machine_id_provider_get_id (id_provider, &unparsed_id_direct, id));
   gchar unparsed_id[HYPHENS_IN_ID + FILE_LENGTH];
   uuid_unparse_lower (id, unparsed_id);
   g_autofree gchar* unhypenated_id = unhyphenate_uuid (unparsed_id);
   g_assert_cmpstr (TESTING_TRACKING_ID, ==, unhypenated_id);
+
+  // adjust \n to match  TESTING_TRACKING_ID
```
wjt: Can we knock the `\n` out of the `TESTING_TRACKING_ID` constant instead? This code is not safe because it's not guaranteed that unparsed_id_direct is at least 34 elements long.

uajain: The reasoning present the test file doesn't apply 100% here (because this is different from /etc/machine-id) but I think we should still respect it?
```
 * The expected size in bytes of the file located at
 * #EmerMachineIdProvider:path.
 * According to http://www.freedesktop.org/software/systemd/man/machine-id.html
 * the file should be 32 lower-case hexadecimal characters followed by a
 * newline character.
```

wjt: My point is that it is not guaranteed that the `unparsed_id_direct` buffer returned is 34 bytes long (though in practice it will be the case), so it's not safe to modify it on the lines below. Another option would be to use:

```c
g_assert_true (strlen (unparsed_id_direct) == 32);
g_assert_cmpmem (TESTING_TRACKING_ID, 32, unparsed_id_direct, 32);
```

or similar.

wjt: What we are trying to test here is that the value returned by `emer_machine_id_provider_get_id()` is what we expect: the 32 hexits from `TESTING_TRACKING_ID`, with no trailing newline, because that's what we want to expose over D-Bus.

uajain: > Another option would be to use:
> 
> ```c
> g_assert_true (strlen (unparsed_id_direct) == 32);
> g_assert_cmpmem (TESTING_TRACKING_ID, 32, unparsed_
> ```

Adopted this approach for now. Addressed, please review. @wjt

wjt: Looks good though I took the liberty of tweaking it slightly. Will merge once CI passes.

PR #238, opened 2020-07-27
daemon: Add tracking-id-file-path as argument

mschiu77:
The eos-event-recorder-daemon can specify the tracking-id file path for each collected metrics data.

https://phabricator.endlessm.com/T30475

### DIFF
--- a/daemon/emer-daemon.h
+++ b/daemon/emer-daemon.h
@@ -73,7 +73,8 @@ struct _EmerDaemonClass
 GType                    emer_daemon_get_type                 (void) G_GNUC_CONST;
 
 EmerDaemon *             emer_daemon_new                      (const gchar             *persistent_cache_directory,
-                                                               EmerPermissionsProvider *permissions_provider);
+                                                               EmerPermissionsProvider *permissions_provider,
+                                                               EmerMachineIdProvider *machine_id_provider);
 
 EmerDaemon *             emer_daemon_new_full                 (GRand                   *rand,
                                                                const gchar             *server_uri,
--- a/daemon/emer-main.c
+++ b/daemon/emer-main.c
@@ -378,6 +378,8 @@ make_daemon (gint                argc,
   g_autofree gchar *persistent_cache_directory = NULL;
   const gchar *persistent_cache_directory_const = PERSISTENT_CACHE_DIR;
   g_autofree gchar *config_file_path = NULL;
+  g_autoptr(EmerMachineIdProvider) machine_id_prov = NULL;
+  g_autofree gchar *tracking_id_file_path = NULL;
   g_autoptr(EmerPermissionsProvider) permissions_provider = NULL;
   GOptionEntry option_entries[] =
   {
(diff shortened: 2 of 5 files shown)


daemon/emer-main.c
```diff
@@ -414,8 +419,13 @@ make_daemon (gint                argc,
     permissions_provider =
       emer_permissions_provider_new_full (config_file_path, NULL);
 
+  if (tracking_id_file_path != NULL)
+    machine_id_prov =
+      emer_machine_id_provider_new_full(tracking_id_file_path);
```
jprvita, line 424: Where is this function implemented?

daemon/emer-daemon.h
```diff
@@ -73,7 +73,8 @@ struct _EmerDaemonClass
 GType                    emer_daemon_get_type                 (void) G_GNUC_CONST;
 
 EmerDaemon *             emer_daemon_new                      (const gchar             *persistent_cache_directory,
-                                                               EmerPermissionsProvider *permissions_provider);
+                                                               EmerPermissionsProvider *permissions_provider,
+                                                               EmerMachineIdProvider *machine_id_provider);
```
jprvita, line 77: Maybe we should use `EmerTrackingIdProvider` and `tracker_id_provider` to name these, since that is the newer name? Same applies to other uses of this name across these changes.

daemon/emer-main.c
```diff
@@ -414,8 +419,13 @@ make_daemon (gint                argc,
     permissions_provider =
       emer_permissions_provider_new_full (config_file_path, NULL);
 
+  if (tracking_id_file_path != NULL)
+    machine_id_prov =
```
jprvita, line 423: Let's maintain the consistency with the rest of the file and spell out "provider" fully, unless you have a good reason to abbreviate it (for example, the longer name making legibility terrible somewhere else).

daemon/emer-main.c
mschiu77: It's in daemon/emer-machine-id-provider.c line 171.

daemon/emer-daemon.h
mschiu77: I reuse the property `tracking-id-path` in EmerMachineIdProvider, so I follow the same naming rule.

daemon/emer-main.c
mschiu77: I referred to `daemon/emer-daemon.c` line 1193. So I think it should be acceptable.
```
set_machine_id_provider (EmerDaemon            *self,
                         EmerMachineIdProvider *machine_id_prov)
```

daemon/emer-main.c
jprvita: Sorry, my bad, I thought you were introducing this new class.

jprvita:
Sorry about my previous overlook that this was not a new class.

The changes look good. There is only one minor nitpick: can you please also update the comments on `emer_machine_id_provider_new_full` implementation, which currently states that it is only useful for unit tests, which is not the case anymore after these changes?

mschiu77:
Thanks. That's the point I didn't notice. I modified the comment. Please help review it. Thanks

jprvita:
Did a small tweak to the comment on `emer_machine_id_provider_new_full` to keep the previous suggested use as well as integrating this new use. Also squashed all commits into one so we don't introduce a point in history that breaks the build.

PR #242, opened 2020-10-15
eos-metrics-collector.exe style trivia

wjt:
I had a quick read of this script and applied some linters.

### DIFF
--- a/tools/eos-metrics-collector.exe
+++ b/tools/eos-metrics-collector.exe
@@ -2,7 +2,6 @@
 
 import argparse
 import dbus
-import errno
 import os
 import shutil
 import subprocess
--- a/tools/eos-metrics-collector.exe
+++ b/tools/eos-metrics-collector.exe
@@ -19,25 +19,25 @@ def get_eos_version():
     for line in f.readlines():
         key, value = line.split("=")
         if key == "ID":
-            distro = value.rstrip('\n').strip('"')
+            distro = value.rstrip("\n").strip('"')
         if key == "VERSION_ID":
-            version = value.rstrip('\n').strip('"')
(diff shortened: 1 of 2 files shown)


tools/eos-metrics-collector.exe
```diff
@@ -48,12 +47,12 @@ class OfflineMetrics():
     def metrics_unit(self, operation):
         # TODO: the metrics daemon is also D-Bus activatable, so on stop we
         # should also do something to prevent it being started over D-Bus.
```
wjt, line 49: Yes, but, I believe it's activated via systemd so masking the systemd unit should DTRT.

tools/eos-metrics-collector.exe
```diff
@@ -2,7 +2,6 @@
 
 import argparse
 import dbus
-import errno
```
manuq, line 5: :+1:

manuq (approved):
Looks good. There are some ways to move forward to keep the repository free of lint errors:
- Run black with a (pre-commit)[https://github.com/psf/black#version-control-integration] hook.
- Setup a python virtualenv with both black and pre-commit as dev requirements. So all devs will have the same environment. [pipenv](https://pypi.org/project/pipenv/), [fades](https://fades.readthedocs.io/en/stable/) and similar tools make this easy.
- Integrate black with the CI tool.

wjt:
I support automation!

jprvita:
In my defense, I brought flake8 with me along this journey (but obviously forgot about it on the last commit where I added `errno` back). I did not know about black.

I like the idea of having a standardized dev environment -- it saves me from having to get a refresher on python's best practices twice a year when I have to touch some python code again :)

PR #243, opened 2020-10-16
collector: unmask the metrics daemon to prevent start over D-Bus

mschiu77:
The metrics daemon also needs to be masked so it won't be started
until the collect/upload is done.

### DIFF
--- a/tools/eos-metrics-collector.exe
+++ b/tools/eos-metrics-collector.exe
@@ -45,14 +45,19 @@ class OfflineMetrics:
             self.metrics_id = self.machine_id_path
 
     def metrics_unit(self, operation):
-        # TODO: the metrics daemon is also D-Bus activatable, so on stop we
-        # should also do something to prevent it being started over D-Bus.
-        cmd = ["systemctl", operation]
-        if operation == "is-active":
-            cmd.append("--quiet")
-        cmd.append(self.systemd_service)
-        returncode = subprocess.call(cmd)
-        return returncode == 0
+        supported_ops = ["mask", "unmask"]
+        if operation not in supported_ops:
+            raise ValueError
+        if operation == "mask":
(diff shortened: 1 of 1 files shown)


tools/eos-metrics-collector.exe
```diff
@@ -45,11 +45,11 @@ class OfflineMetrics:
...
-        # TODO: the metrics daemon is also D-Bus activatable, so on stop we
-        # should also do something to prevent it being started over D-Bus.
         cmd = ["systemctl", operation]
         if operation == "is-active":
             cmd.append("--quiet")
+        elif operation == "mask":
+            cmd.append("--now")
         cmd.append(self.systemd_service)
         p = subprocess.run(cmd)
         return p.returncode == 0
```
jprvita, line 55:
```suggestion
        supported_ops = ["mask", "unmask"]                                                                             
        if operation not in supported_ops:
            raise ValueError
        if operation == "mask":
            cmd = ["systemctl", "mask", "--runtime", "--now", "--quiet"]
            cmd.append(self.systemd_service)
            subprocess.call(cmd)
        else
            systemd_runtime_dir = "/run/systemd/system"
            try:
                os.remove(os.path.join(systemd_runtime_dir, self.systemd_service))
            except FileNotFoundError:
                pass
```

jprvita (changes requested):
I left a longer rationale on the ticket, but I think the best solution here is to stop+runtime-mask the unconditionally unmask the unit.

The `metrics_unit` implementation is suggested here, but you'll need to adjust the calls to it accordingly.

(also, please remember to add the phabricator link to the commit message and PR description)

jprvita:
Force-pushed to fix a missing blank line (there should be two blank lines before a new class definition) and rebased to master.

PR #244, opened 2020-10-16
T30974

mschiu77:
Fix the compatibility problem caused by the f"" string usage and subprocess.run

### DIFF
--- a/tools/eos-metrics-collector.exe
+++ b/tools/eos-metrics-collector.exe
@@ -51,8 +51,8 @@ class OfflineMetrics:
         if operation == "is-active":
             cmd.append("--quiet")
         cmd.append(self.systemd_service)
-        p = subprocess.run(cmd)
-        return p.returncode == 0
+        returncode = subprocess.call(cmd)
+        return returncode == 0
 
 
 class OfflineMetricsUploader(OfflineMetrics):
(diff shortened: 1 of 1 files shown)


jprvita:
- Please split the mask / unmask changes into its own separate commit -- having each commit implementing only one thing (as much as possible) makes the review process a lot easier. And on those changes, let's not change the `metrics_unit` API -- let's keep the callers passing "start" or "stop", and encapsulate the extra necessary steps in `metrics_unit`'s implementation.
- The changes from f-strings to .format look good, except that they introduce some long lines that are likely to break the coding style. Please run each commit through [black](https://github.com/psf/black) and clean any warnings.
- The changes from `.run` to `.call` look good to me, but I haven't tested them yet.
- Finally, please make sure to always add a link to the phabricator task on each commit message, and on the PR description, so our phabricator integrations work correctly.

jprvita (changes requested):
- Please split the mask / unmask changes into its own separate commit -- having each commit implementing only one thing (as much as possible) makes the review process a lot easier. And on those changes, let's not change the `metrics_unit` API -- let's keep the callers passing "start" or "stop", and encapsulate the extra necessary steps in `metrics_unit`'s implementation.
- The changes from f-strings to .format look good, except that they introduce some long lines that are likely to break the coding style. Please run each commit through [black](https://github.com/psf/black) and clean any warnings.
- The changes from `.run` to `.call` look good to me, but I haven't tested them yet.
- Finally, please make sure to always add a link to the phabricator task on each commit message, and on the PR description, so our phabricator integrations work correctly.

mschiu77:
I do the mask/unmask thing in another T30975. I didn't notice that it's also in this this PR. Thanks for pointing that out.

tools/eos-metrics-collector.exe
```diff
@@ -234,12 +235,12 @@ def collect(storage_root):
     data_size = collector.calculate_metrics_data_size()
     if free_space < data_size:
         msg = (
-            f"Insufficient free space ({free_space} bytes) for metrics "
-            f"data size ({data_size} bytes), aborting."
+            "Insufficient free space ({} bytes) for ".format(free_space) +
+            "metrics data size ({} bytes), aborting.".format(data_size)
```
jprvita, line 239:
```suggestion
            "Insufficient free space ({} bytes) for metrics data size "
            "({} bytes), aborting.".format(free_space, data_size)
```

jprvita (changes requested):
I think you missed what black suggested -- when I run it on your latest branch it still changed those strings you are concatenating with `+`. But I suggest getting rid of the `+` altogether and only calling `.format()` once per string, as in the inline suggestions here.

But please apply them directly to the commit introducing the change (the f-string one) instead of accepting them in github, which I think will create a fixup commit on the tip of the branch instead. This keep the history a bit cleaner.

jprvita (approved):

PR #247, opened 2020-12-22
collector: force env to production while doing upload

mschiu77:
To ensure the collected data are submitted to the correct metrics
server.

https://phabricator.endlessm.com/T31220

### DIFF
--- a/tools/eos-metrics-collector.exe
+++ b/tools/eos-metrics-collector.exe
@@ -60,6 +60,7 @@ class OfflineMetricsUploader(OfflineMetrics):
         self.bin_name = "eos-metrics-event-recorder"
         self.bin_path = os.path.join(self.bin_dir, self.bin_name)
         self.up_trigger_bin = "eos-upload-metrics"
+        self.metrics_env_bin = "eos-select-metrics-env"
         self.tmpdir = ""
         self.daemon = None
 
@@ -70,6 +71,11 @@ class OfflineMetricsUploader(OfflineMetrics):
         if os.path.exists(self.tmpdir):
             shutil.rmtree(self.tmpdir)
         self.metrics_unit("unmask")
+
+    def set_metrics_production_env(self):
+        return subprocess.call(
+            self.metrics_env_bin, "production", stdout=subprocess.DEVNULL
+        )
 
     def chusr(self):
         pwd_entry = pwd.getpwnam("metrics")
@@ -122,6 +128,7 @@ class OfflineMetricsUploader(OfflineMetrics):
             sys.exit(0)
 
         self.metrics_unit("mask")
+        self.set_metrics_production_env()
 
         self.tmpdir = tempfile.mkdtemp()
         shutil.chown(self.tmpdir, user="metrics", group="metrics")


jprvita:
Merged, with the following changes:

1. Style change, suggested by black:
```
diff --git a/tools/eos-metrics-collector.exe b/tools/eos-metrics-collector.exe
index 1ff4c54..d70b932 100755
--- a/tools/eos-metrics-collector.exe
+++ b/tools/eos-metrics-collector.exe
@@ -73,7 +73,9 @@ class OfflineMetricsUploader(OfflineMetrics):
         self.metrics_unit("unmask")
 
     def set_metrics_production_env(self):
-        return subprocess.call(self.metrics_env_bin, "production", stdout=subprocess.DEVNULL)
+        return subprocess.call(
+            self.metrics_env_bin, "production", stdout=subprocess.DEVNULL
+        )
 
     def chusr(self):
         pwd_entry = pwd.getpwnam("metrics")
```
2. It does not make any difference in practice, but semantically, it makes more sense to first stop+mask the unit, then to modify its configuration. This is especially true if we decide to restore the previous configuration, but I don't think that is necessary atm.
```
diff --git a/tools/eos-metrics-collector.exe b/tools/eos-metrics-collector.exe
index 1ff4c54..d70b932 100755
--- a/tools/eos-metrics-collector.exe
+++ b/tools/eos-metrics-collector.exe
@@ -124,8 +126,9 @@ class OfflineMetricsUploader(OfflineMetrics):
             msg = "No metrics data found in {}, nothing to do.".format(self.storage_dir)
             print(msg)
             sys.exit(0)
-        self.set_metrics_production_env()
+
         self.metrics_unit("mask")
+        self.set_metrics_production_env()
 
         self.tmpdir = tempfile.mkdtemp()
         shutil.chown(self.tmpdir, user="metrics", group="metrics")
```

wjt:
This won't work if the OStree URL points at the staging repo because the Daemon reads it at upload time and changes the metrics env back if needed. So we still need to document (or fix) that behaviour

jprvita:
> This won't work if the OStree URL points at the staging repo because the Daemon reads it at upload time and changes the metrics env back if needed. So we still need to document (or fix) that behaviour

@wjt Do we expect end-users to have their machines pointing at the ostree staging repo? (I'm not sure how this works for beta users) And moreover, do we expect the metrics server to be configured to anything other than production on end-users' systems? If so, in which situations?

wjt:
My point is not so much that we *expect* many users to be pointing at the staging repo (though yes, this includes beta users).

My point is that if the script is run on a machine that's pointed at the staging repo (either dev or beta) – perhaps a developer is helping a partner upload metrics – then the script will happily upload the metrics to the development metrics environment, then delete the data from the USB stick. Then they would effectively be lost without some manual intervention to try to dig all the events out of the dev metrics instance and manually ingest them into the prod one.

I think the script should refuse to run if the ostree remote is pointed at the staging repo; or we need to change the daemon so that we can guarantee that metrics uploaded by this mechanism are uploaded to a particular environment.

jprvita:
> My point is not so much that we _expect_ many users to be pointing at the staging repo (though yes, this includes beta users).
> 
> My point is that if the script is run on a machine that's pointed at the staging repo (either dev or beta) – perhaps a developer is helping a partner upload metrics – then the script will happily upload the metrics to the development metrics environment, then delete the data from the USB stick. Then they would effectively be lost without some manual intervention to try to dig all the events out of the dev metrics instance and manually ingest them into the prod one.
> 
> I think the script should refuse to run if the ostree remote is pointed at the staging repo; or we need to change the daemon so that we can guarantee that metrics uploaded by this mechanism are uploaded to a particular environment.

Ok, makes sense, I'll follow-up on the ticket.

PR #249, opened 2021-01-25
Allow configuring the default metrics server at build time

mcatanzaro:
This makes it easy to use eos-event-recorder-daemon with a metrics server other than metrics.endlessm.com.

I'm not sure whether metrics.endlessm.com is really a good default. It might make more sense to force users to explicitly configure the server to use and then update the eos packaging to select metrics.endlessm.com, to ensure nobody accidentally uses the production server.

https://phabricator.endlessm.com/T31278

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -41,10 +41,11 @@ AM_CFLAGS = @WARN_CFLAGS@
 AM_LDFLAGS = @WARN_LDFLAGS@
 
 # Enable strict compiler flags and install systemd unit files to the specified
-# directory when doing 'make distcheck'.
+# directory when doing 'make distcheck'. Set a dummy default server URL.
 AM_DISTCHECK_CONFIGURE_FLAGS = \
 	--enable-compile-warnings=error \
 	--with-systemdsystemunitdir='$${prefix}/lib/systemd/system' \
+	--with-default-metrics-server=azafea.example \
 	$(NULL)
 if EOS_ENABLE_COVERAGE
 AM_DISTCHECK_CONFIGURE_FLAGS += --enable-coverage --with-coverage-dir=@EOS_COVERAGE_DIR@
--- a/configure.ac
+++ b/configure.ac
@@ -128,6 +128,12 @@ AS_IF([test -f $EVENT_RECORDER_SERVER_XML],
     [AC_MSG_RESULT([present])],
     [AC_MSG_ERROR([absent])])
 
+AC_ARG_WITH([default_metrics_server],
+    [AS_HELP_STRING([--with-default-metrics-server=HOSTNAME],[choose default metrics server])],
+    [DEFAULT_METRICS_SERVER="\"$withval\""],
+    [AC_MSG_ERROR([You must specify a default metrics server using --with-default-metrics-server])])
+AC_DEFINE_UNQUOTED([DEFAULT_METRICS_SERVER], [$DEFAULT_METRICS_SERVER], [Default metrics server])
+
 # Required libraries
 # ------------------
 PKG_CHECK_MODULES([EOS_EVENT_RECORDER_DAEMON], [$EMER_REQUIRED_MODULES])
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -20,6 +20,7 @@
  * <http://www.gnu.org/licenses/>.
  */
 
+#include "config.h"
 #include "emer-daemon.h"
 
 #include <math.h>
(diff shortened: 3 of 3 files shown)


wjt:
@eos-shell-build OK to test

> I'm not sure whether metrics.endlessm.com is really a good default. It might make more sense to force users to explicitly configure the server to use and then update the eos packaging to select metrics.endlessm.com, to ensure nobody accidentally uses the production server.

I agree with this. If you're happy to make that change (or even just the configure.ac half of it) that'd be great.

You'll have seen this reference to there being 3 different "environments" – prod, dev, test – but you may not be aware of the logic that [reads the URL for the `eos` ostree remote](https://github.com/endlessm/eos-event-recorder-daemon/blob/77db80a556c32e8af6f07045e8e217ca667c3dbd/daemon/emer-permissions-provider.c#L181-L273) and [changes the environment from prod to dev if the URL contains `staging`](https://github.com/endlessm/eos-event-recorder-daemon/blob/77db80a556c32e8af6f07045e8e217ca667c3dbd/daemon/emer-permissions-provider.c#L657-L668). For a non-eos distro you may want to poke at this code a bit because it will issue a bunch of `g_warning`s on non-eos systems.

mcatanzaro:
> I agree with this. If you're happy to make that change (or even just the configure.ac half of it) that'd be great.

I will update this PR accordingly, and I'll submit a second PR for the -debian branch to update the rules file. But can I rely on you to test the -debian PR? It's been a long time since I last built a Debian package, and I would be embarrassed if I break it with some silly typo. :P

> You'll have seen this reference to there being 3 different "environments" – prod, dev, test – but you may not be aware of the logic that [reads the URL for the `eos` ostree remote](https://github.com/endlessm/eos-event-recorder-daemon/blob/77db80a556c32e8af6f07045e8e217ca667c3dbd/daemon/emer-permissions-provider.c#L181-L273) and [changes the environment from prod to dev if the URL contains `staging`](https://github.com/endlessm/eos-event-recorder-daemon/blob/77db80a556c32e8af6f07045e8e217ca667c3dbd/daemon/emer-permissions-provider.c#L657-L668). For a non-eos distro you may want to poke at this code a bit because it will issue a bunch of `g_warning`s on non-eos systems.

I hadn't noticed that particular spot yet, but there are going to be many other such places where eos-specific assumptions would need to be relaxed when not running on eos. Anyway, I will add a TODO.

mcatanzaro:
Since

mcatanzaro:
OK, I've updated this PR to fail the build if the default metrics server is not configured. I *expect* that's why the CI is failing. I've created #250. Hopefully that should be enough to fix the CI? I think it should be safe to merge that first, because an unrecognized configure argument should be ignored.

wjt:
@eos-shell-build  retest this please

wjt:
Sigh. It's not as simple as that because Jenkins doesn't use the flags from the Debian packaging, it just runs configure. Those flags are only used once the package makes it through the gauntlet into OBS.

Some ideas:

1. I can set the default server in the Jenkins configuration. A little annoying for folks hacking on this but they can easily do the same.
2. The option can default to (e.g.) `localhost` and warn rather than failing (though I can see why you didn't do this).

mcatanzaro:
> * I can set the default server in the Jenkins configuration. A little annoying for folks hacking on this but they can easily do the same.

That sounds fine.

> * The option can default to (e.g.) `localhost` and warn rather than failing (though I can see why you didn't do this).

Ehh... nobody looks at configure warnings, so that doesn't seem great. I can implement this if you want, though. (Or just go back to defaulting to metrics.endlessm.com.)

dbnicholson:
> > I hadn't noticed that particular spot yet
> 
> Actually it's causing this warning that I already had on my TODO:
> 
> ```
> Jan 26 11:08:46 rhel8 eos-metrics-eve[1218]: Unable to open OSTree repo. Error: opening repo: opendir(/ostree/repo): No such file or director>
> ```
> 
> I think we can either (a) check /etc/os-release and only run that code if the system is Endless, or (b) silently fail and move on if /ostree does not exist. Either way, it can be handled in a separate PR.

Probably best is to check if `ostree=` is in the kernel command line to see if it's an ostree booted system. That's what ostree itself does and I believe there's libostree API for it (can't recall at the moment).

wjt:
retest this please

wjt:
retest this please

wjt:
retest this please

I had to push an additional change to pass a dummy URL during `make distcheck`.

PR #254, opened 2021-08-18
daemon: Stop creating tracking-id

starnight:
According to new metrics protocol, tracking-id is not used any more.
However, to solve the problem of uploading buffered events from older
EOS version like 3.9 and earlier, we should keep some features of
machine-id-provider, but do not create tracking-id file if it does not
exsits.

https://phabricator.endlessm.com/T32454

### DIFF
--- a/daemon/emer-main.c
+++ b/daemon/emer-main.c
@@ -288,11 +288,6 @@ on_bus_acquired (GDBusConnection *system_bus,
 {
   EmerDaemon *daemon = EMER_DAEMON (user_data);
   EmerEventRecorderServer *server = emer_event_recorder_server_skeleton_new ();
-  g_autofree gchar *tracking_id = NULL;
-
-  tracking_id = emer_daemon_get_tracking_id (daemon);
-  emer_event_recorder_server_set_tracking_id (server,
-                                              tracking_id != NULL ? tracking_id : "");
 
   g_signal_connect (server, "handle-record-singular-event",
                     G_CALLBACK (on_record_singular_event), daemon);
--- a/daemon/emer-main.c
+++ b/daemon/emer-main.c
@@ -102,28 +102,6 @@ on_set_enabled (EmerEventRecorderServer *server,
   return TRUE;
 }
 
-static gboolean
-on_reset_tracking_id (EmerEventRecorderServer *server,
-                      GDBusMethodInvocation   *invocation,
-                      EmerDaemon              *daemon)
-{
-  g_autoptr(GError) error = NULL;
-  g_autofree gchar *tracking_id = NULL;
-
-  if (!emer_daemon_reset_tracking_id (daemon, &error))
-    {
-      g_dbus_method_invocation_return_gerror (invocation, error);
-      return TRUE;
-    }
-
-  tracking_id = emer_daemon_get_tracking_id (daemon);
-  emer_event_recorder_server_set_tracking_id (server,
-                                              tracking_id != NULL ? tracking_id : "");
-
-  emer_event_recorder_server_complete_reset_tracking_id (server, invocation);
-  return TRUE;
-}
-
 static void
 handle_upload_finished (EmerDaemon       *daemon,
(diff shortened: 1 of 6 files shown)


daemon/emer-main.c
```diff
@@ -291,8 +291,6 @@ on_bus_acquired (GDBusConnection *system_bus,
   g_autofree gchar *tracking_id = NULL;
 
   tracking_id = emer_daemon_get_tracking_id (daemon);
```
wjt, line 293: These are now unused and can be removed.

tests/daemon/test-daemon.c
```diff
@@ -1402,48 +1401,15 @@ test_daemon_refresh_and_reload_machine_id (Fixture       *fixture,
...
-                                   new_machine_id);
-
-  g_assert (uuid_compare (initial_machine_id, new_machine_id) != 0);
-
-  /* Send some metrics using the newly created daemon. We'll assert
-   * one we receive the metrics that the received metric has the same
-   * machine-id as the one provided by the machine-id provider */
+  /* Send some metrics using the created daemon. We'll assert one we receive the
+   * metrics that the received metric has the same machine-id as the one
+   * provided by the machine-id provider */
```
wjt, line 1406: This message is no longer true. We don't make any assertion about the machine ID. Is this test redundant with another test which calls `record_singulars`, `assert_singulars_received`, `wait_for_upload_to_finish`?

wjt (changes requested):

daemon/emer-main.c
starnight: ok

tests/daemon/test-daemon.c
starnight: You are right!  This becomes redundant.

tests/daemon/test-machine-id-provider.c
```diff
@@ -115,33 +115,6 @@ teardown (MachineIdTestFixture *fixture,
...
-  g_assert_false (g_file_test (fixture->tracking_id_file_path, G_FILE_TEST_EXISTS));
-
-  g_autoptr(EmerMachineIdProvider) id_provider =
-    emer_machine_id_provider_new_full (fixture->tracking_id_file_path);
-
-  uuid_t id;
-  // id_provider_get_id will write a new tracking ID, if no ID is found.
-  g_assert (emer_machine_id_provider_get_id (id_provider, NULL, id));
-
-  g_assert (g_file_test (fixture->tracking_id_file_path, G_FILE_TEST_EXISTS));
```
wjt, line 134: I missed this last time but perhaps we should keep this test case, but adjust it to assert that get_id fails if the file doesn't exist, and g_file_test (..., EXISTS) is still false after the call. This will exercise the "file doesn't exist" path.

wjt (approved):
LGTM with the exception of that test suggestion – up to you if you want to resurrect the test and adjust it, or not bother.

tests/daemon/test-machine-id-provider.c
starnight: Let's check failing to get the tracking-id.

tests/daemon/test-machine-id-provider.c
```diff
@@ -116,8 +116,8 @@ teardown (MachineIdTestFixture *fixture,
 // Testing Cases
 
 static void
-test_machine_id_provider_create_tracking_id_if_unavailable (MachineIdTestFixture *fixture,
-                                                            gconstpointer         tdata)
+test_machine_id_provider_have_no_tracking_id_if_unavailable (MachineIdTestFixture *fixture,
+                                                             gconstpointer         tdata)
 {
   g_autoptr(GError) error = NULL;
```
wjt, line 122: Unusued

wjt: So is the one below

starnight: Ah!  Thank you!

PR #255, opened 2021-08-26
Aggregate timers

GeorgesStavracas:
Start implementing it on top of Jian-Hong's work.

To Do:

 * [x] Save these events in an on-disk buffer
 * [x] Submit them to the server

https://phabricator.endlessm.com/T32203

### DIFF
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -256,7 +256,7 @@ buffer_event (EmerDaemon *self,
       if (event_cost > MAX_REQUEST_PAYLOAD)
         {
           g_warning ("Dropping %" G_GSIZE_FORMAT "-byte event. The maximum "
-                     "permissable event size (including type string with "
+                     "permissible event size (including type string with "
                      "null-terminating byte) is %d bytes.",
                      event_cost, MAX_REQUEST_PAYLOAD);
         }


daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+  filename = g_strdup_printf ("%s-%u-%s",
+                              aggregate_key_str,
+                              timer_data->unix_user_id,
+                              event_id_str);
+
+  return g_build_filename (priv->persistent_cache_directory,
+                           "aggregate-timers",
+                           date_string,
+                           filename,
+                           NULL);
```
dbnicholson, line 1196: It's probably fine to have a file per aggregate key per user per event per day. I'd think some sort of indexed file/database would be nicer to use, particularly if there's a single reader and writer. I don't have any great ideas there, though.

dbnicholson (commented):
All seems pretty reasonable to me. I'm interested to see the last couple pieces fleshed out:

* Handling the monthly aggregation
* Clearing queued tallies
* Handling metrics disabled and permissions changes

daemon/emer-session.c
```diff
@@ -0,0 +1,120 @@
...
+    g_warning ("Wrong event ID format");
+    return NULL;
+  }
+
+  UserSesstion *session = g_new0 (UserSesstion, 1);
+  session->start_time = g_date_time_new_now_local ();
+  session->start_mclock = g_get_monotonic_time ();
+  session->user_id = user_id;
+  if (!g_strncasecmp (eid, DESKTOP_APPLICATION_EVENT_ID, UUID_LENGTH))
+    sessions->desktop_file_id = /* desktop file ID */;
```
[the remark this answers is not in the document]
wjt: I don't think the daemon should have particular knowledge of the event ID for this UUID. Why does it need that? Can't it use the UUID itself as the filename, and store the variant payload as-is?

daemon/emer-daemon.c
```diff
@@ -1665,3 +2072,67 @@ emer_daemon_get_permissions_provider (EmerDaemon *self)
...
+                    timer_data);
+
+  g_hash_table_insert (priv->aggregate_timers,
+                       g_variant_ref (cache_key),
+                       timer_data);
+
+  if (priv->dispatch_aggregate_timers_daily_source_id == 0)
+    schedule_next_midnight_tick (self);
+  if (priv->dispatch_aggregate_timers_monthly_source_id == 0)
+    schedule_next_month_tick (self);
```
wjt, line 2137: So here we are only scheduling these ticks if at least 1 timer is running. I think it might be clearer to make them unconditional while the daemon is running. On start-up it needs to check for missed ticks anyway.

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+static gboolean
+on_timer_stopped_cb (EmerAggregateTimer    *timer,
+                     GDBusMethodInvocation *invocation,
+                     AggregateTimerData    *timer_data)
+{
+  EmerDaemon *self = timer_data->daemon;
+  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
+
+  // Removing from the hash table will call stop_aggregate_timer()
+  g_hash_table_remove (priv->aggregate_timers, timer_data->cache_key);
```
wjt, line 1495: I believe that after this point 'timer_data' will have been freed by the hash table? Perhaps worth a note to that effect.

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+  g_clear_pointer (&timer_data->payload, g_variant_unref);
+  g_free (timer_data);
+}
+
+static gboolean
+on_timer_stopped_cb (EmerAggregateTimer    *timer,
+                     GDBusMethodInvocation *invocation,
+                     AggregateTimerData    *timer_data)
+{
+  EmerDaemon *self = timer_data->daemon;
```
wjt, line 1491: Perhaps worth a `EMER_IS_DAEMON` check before dereferencing this?

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+          continue;
+        }
+
+      variant = g_variant_parse (NULL, contents, NULL, NULL, &local_error);
+      if (local_error)
+        {
+          g_warning ("Error parsing tally from file %s: %s",
+                     path, local_error->message);
+          continue;
+        }
```
wjt, line 1373: At some point we should presumably try to delete broken files in this directory to stop them accumulating.

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+      g_file_get_contents (path, &contents, NULL, &local_error);
+
+      if (local_error)
+        {
+          g_warning ("Error loading tally from file %s: %s",
+                     path, local_error->message);
+          continue;
+        }
+
+      variant = g_variant_parse (NULL, contents, NULL, NULL, &local_error);
```
wjt, line 1367: I think this is using the text serialization of the variant. I think we should use the binary serialization. g_variant_new_from_data / g_variant_get_data() + g_variant_get_size()

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+      if (local_error)
+        {
+          g_warning ("Error parsing tally from file %s: %s",
+                     path, local_error->message);
+          continue;
+        }
+
+      g_variant_get (variant,
+                     "(vvvuubmv)",
```
wjt, line 1376: This function has undefined behaviour if the type given does not match the actual type of `variant`. We are reading potentially-untrusted data from disk. A bit-flip may have messed up the on-disk type. We should perform a type check.

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+                 "(vvvuubmv)",
+                 &event_id,
+                 &aggregate_key,
+                 &cache_key,
+                 &unix_user_id,
+                 &previous_counter,
+                 &complete,
+                 &payload);
+
+  g_assert (complete);
```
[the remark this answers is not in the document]
wjt: I don't like calling `g_assert()` on data that has been read from disk.

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+  gboolean complete = timer_data->end_monotonic_us != -1;
+  guint32 counter;
+  gint64 end_time;
+
+  counter = previous_counter;
+
+  end_time = complete ? timer_data->end_monotonic_us : g_get_monotonic_time ();
+  counter += (guint32) end_time - timer_data->start_monotonic_us;
+
+  return g_variant_new ("(vvvuubmv)",
```
wjt, line 1161: If there is documentation about what the fields of this structure are, I can't see it. What is the definition of 'complete' for example?

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+  g_autofree gchar *contents = NULL;
+  g_autoptr(GVariant) variant = NULL;
+  g_autoptr(GVariant) cache_key = NULL;
+  g_autoptr(GVariant) event_id = NULL;
+  g_autoptr(GVariant) aggregate_key = NULL;
+  g_autoptr(GVariant) payload = NULL;
+  g_autoptr(GError) error = NULL;
+  gboolean complete;
+  guint32 previous_counter;
+  uid_t unix_user_id;
```
wjt, line 1212: In practice it is true that `uid_t` is an alias for `uint32_t` with linux+glibc, but I don't think it is guaranteed. Since the D-Bus method GetConnectionUnixUser is guaranteed to return a uint32, and this gvariant is also using the `u` type, maybe we just use guint32 throughout? I don't think this code actually needs to care in particular that it is a uid_t.

daemon/emer-daemon.c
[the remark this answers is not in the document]
wjt: and, if this field is always true, then why is it present? I don't quite understand its semantics.

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+{
+  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
+  AggregateTimerData *timer_data;
+  GHashTableIter iter;
+
+  g_hash_table_iter_init (&iter, priv->aggregate_timers);
+  while (g_hash_table_iter_next (&iter, NULL, (gpointer *) &timer_data))
+    {
+      save_aggregate_timer_to_tally (timer_data);
+      timer_data->start_monotonic_us = g_get_monotonic_time ();
```
wjt, line 1288: Suppose there are 100 running timers. We will call g_get_monotonic_time() 200 times here – once in aggregate_timer_to_variant (via save_aggregate_timer_to_tally) and again here, for each event. Of course it is likely not to have any significant impact on the time in seconds but it would be preferable IMO to use the same monotonic time when cutting the timer in two, as the end of the first part and the start of the second part, and having done that we could use the same monotonic time for all of them.

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+                     &event_id,
+                     &aggregate_key,
+                     &cache_key,
+                     &unix_user_id,
+                     &counter,
+                     &complete,
+                     &payload);
+
+      if (!complete)
+        continue;
```
wjt, line 1386: I don't think we need this field. We know that if the date is before today (this month for the monthly version), the event is complete and can be submitted to the server; and if it's today (or further into the future), it's not.

daemon/emer-daemon.c
```diff
@@ -1123,6 +1142,379 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+    }
+}
+
+static gboolean
+clock_ticked_midnight_cb (gpointer user_data)
+{
+  EmerDaemon *self = user_data;
+
+  save_aggregate_timers_to_tally (self);
+  buffer_daily_aggregate_events_to_queue (self);
```
wjt, line 1402: We also need to "catch up" for any day(s) missed while the computer was switched off.

Suppose I switch off my computer at 6pm on August 31st, and turn it on again at 8am on August 1st. Then we will have missed both the midnight and end-of-month ticks.

wjt (changes requested):
This is shaping up nicely.

On top of the pieces Dan mentioned and my comments below: have you given any thought to unit-testing this? Of course we can't make the unit tests run overnight so we will need to mock out the timers in some way, so that the tests can cause midnight to happen.

In eos-payg there is a clock abstraction, EpgClock, with two implementations, EpgRealClock and EpgFakeClock. Maybe a similar arrangement could work here.

daemon/emer-daemon.c
GeorgesStavracas: What would happen on timezone or wallclock changes? The reason I kept it as a boolean field was to not implode on time travels backwards in time.

daemon/emer-daemon.c
GeorgesStavracas: As you wrote on https://phabricator.endlessm.com/T32203:

> At midnight², e-e-r-d adds any running timers to the corresponding tally [...] and then adds what I'll call **"complete"** aggregate events based on the data in the tally

Which I interpreted as aggregate events that had their timer stopped.

daemon/emer-daemon.c
GeorgesStavracas: Added a call to `g_variant_check_format_string` (I'll switch the variant type string into a define)

daemon/emer-daemon.c
wjt: We could instead delete data from this cache once it is enqueued for submission to the server. This would risk double-counting users if the clock moves backwards to the previous day but avoids the risk of double-counting time, I think.

daemon/emer-daemon.c
wjt: I didn't know about that function! I have always extracted the type and compared it as a separate step. Nice.

daemon/emer-daemon.c
wjt: Thinking about it a bit more: at midnight (resp end-of-month) we iterate across all running timers, add the time elapsed so far to the on-disk tally, and reset their start time. Only after that do we submit to the server. So I think we can say that any data that is in the on-disk tally at the point where we iterate across it all is "complete" in this sense.

wjt:
I had a think overnight about testing, and about this Slack comment:

> I'm seriously considering moving all this code into a separate file, but it's just so intertwined with the daemon itself it might be hard 

Here's an idea of how it could be structured:

- Represent the tally with an EmerAggregateTally class, with a function to increment a tally for a particular date, event UUID, and other keys, and another function to iterate across all tallies, returning the keys and values. No timeouts or D-Bus stuff in here.
- Represent running timers with their own class, EmerAggregateTimer. Don't put any D-Bus stuff in here – give it a `emer_aggregate_timer_stop()` function, and a `split()` function (name TBD) that will be called to reset the start time at midnight. Rather than calling `g_get_monotonic_time()`, pass the monotonic time as an argument to each function.

I think the EmerAggregateTimer could probably be given the Tally as a construct-time parameter, then `stop()` and `split()` can update the tally.

Now they can easily be tested in isolation without having to think about clocks and D-Bus.

emer-daemon would own an EmerAggregateTally and list of EmerAggregateTimers, as well as having the timeouts to split and submit data. (Unclear whether the D-Bus bit belongs in -daemon.c or -main.c – the former is easier but the latter is more consistent with the current structure.)

daemon/emer-daemon.c
```diff
@@ -1665,3 +1875,63 @@ emer_daemon_get_permissions_provider (EmerDaemon *self)
   EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
   return priv->permissions_provider;
 }
+
+void
+emer_daemon_start_aggregate_timer (EmerDaemon            *self,
+                                   GDBusMethodInvocation *invocation,
```
wjt, line 1881: Rather than passing the invocation in here, can you make this return the timer on success and set a `GError **` on failure? There's no other DBus logic in this source file.

daemon/emer-daemon.c
```diff
@@ -1123,6 +1145,169 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+
+  save_aggregate_timers_to_tally (self, date);
+  buffer_daily_aggregate_events_to_queue (self);
+  schedule_next_midnight_tick (self);
+
+  return G_SOURCE_REMOVE;
+}
+
+static gboolean
+clock_turned_month_cb (gpointer user_data)
```
wjt, line 1227: I don't think you need two timers. You can update the current/just-ended month's tallies at the same time as the midnight tick

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,373 @@
...
+                                                    unix_user_id,
+                                                    event_id,
+                                                    aggregate_key,
+                                                    date);
+
+  if (!g_file_test (tally_path, G_FILE_TEST_EXISTS))
+    return 0;
+
+  g_file_get_contents (tally_path, &contents, &size, &error);
+  if (error)
```
wjt, line 113: Personally I prefer this style, but I accept it is a matter of preference :)

```suggestion
  if (!g_file_get_contents (tally_path, &contents, &size, &error))
```

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,373 @@
...
+    return 0;
+
+  variant = g_variant_new_from_data (G_VARIANT_TYPE ("(vvuumv)"),
+                                     contents, size, FALSE, NULL, NULL);
+  if (error)
```
wjt, line 118: `error` isn't passed to `g_variant_new_from_data` so this error path will never be taken. I actually don't know what happens if you pass total junk to `g_variant_new_from_data` but I presume that the subsequent `g_variant_check_format_string` is what will fail.

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,373 @@
...
+    return 0;
+
+  if (!g_variant_check_format_string (variant, "(vvuumv)", FALSE))
+    return 0;
+
+  g_variant_get (variant,
+                 "(vvvuumv)",
```
wjt, line 125: This format string is wrong, it has too many `v`s. Perhaps put a `#define` somewhere?

You can either pass `NULL` to ignore a value, or use `g_variant_get_child` to pick out a single field by index. No need for the `unused_*` fields.

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,373 @@
...
+  GObjectClass *object_class = G_OBJECT_CLASS (klass);
+
+  object_class->finalize = emer_aggregate_tally_finalize;
+  object_class->set_property = emer_aggregate_tally_set_property;
+
+  /*
+   * EmerDaemon:persistent-cache-directory:
+   *
+   * A directory for temporarily storing events until they are uploaded to the
+   * metrics servers. If a network send provider is not specified, a default
```
wjt, line 177: "network send provider" – copy-pasta error?

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,373 @@
...
+      result = func (unix_user_id, event_id,
+                     aggregate_key, payload,
+                     counter, date, user_data);
+
+      if (result & EMER_TALLY_ITER_DELETE)
+        {
+          g_autoptr(GError) deletion_error = NULL;
+
+          g_warning ("Error parsing tally from file %s: %s",
+                     path, local_error->message);
```
wjt, line 362: `local_error` may be NULL here. I think this block is copy-pasta from higher up.

daemon/emer-aggregate-tally.c
wjt: I spoke to @pwithnall on `#gnome-hackers` about this. My understanding is that this is expected behaviour.

- check_format_string() is redundant because that is true by construction.
- we can use `g_variant_is_normal_form()` to validate the parsed variant – if non-normal form then it is corrupt
- in this particular function the handling is the same either way – return 0
- but in the other function which loads these, the handling may be different

daemon/emer-daemon.c
```diff
@@ -1123,6 +1145,169 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+
+  emer_daemon_record_aggregate_event (self,
+                                      event_id,
+                                      date,
+                                      counter,
+                                      payload != NULL,
+                                      payload);
+
+  return EMER_TALLY_ITER_CONTINUE;
+}
```
wjt, line 1192: We had discussed not doing this step (which transfers data from the on-disk tally into the possibly in-memory submission queue) and instead reworking the code that reads from the existing submission queue to also pull from these on-disk tallies. Either one seems workable but this way is less durable.

If doing it this way,  wouldn't we want to DELETE here, once the event is in the submission queue?

wjt (commented):
Shaping up! Here is some more feedback in case it's useful

daemon/emer-aggregate-tally.c
GeorgesStavracas: In my defense, EmerDaemon also has this copy-pasta!

daemon/emer-aggregate-tally.c
GeorgesStavracas: TIL, thanks

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,373 @@
...
+          g_warning ("Error loading tally from file %s: %s",
+                     path, local_error->message);
+          continue;
+        }
+
+      variant = g_variant_new_from_data (G_VARIANT_TYPE ("(vvuumv)"),
+                                         contents, size,
+                                         FALSE,
+                                         NULL, NULL);
+      if (local_error)
```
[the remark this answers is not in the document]
GeorgesStavracas: Done

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,391 @@
...
+  g_autoptr(GVariant) variant = NULL;
+  guint32 previous_counter = 0;
+
+  tally_path = get_tally_path_from_aggregate_timer (self,
+                                                    unix_user_id,
+                                                    event_id,
+                                                    aggregate_key,
+                                                    date);
+
+  variant = load_aggregate_timer_from_path (tally_path);
```
GeorgesStavracas, line 164: This works, but it loads the entire file from disk just to retrieve a counter. I wonder if it's possible to not load it all?

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,391 @@
...
+    return NULL;
+
+  g_file_get_contents (path, &contents, &size, &error);
+  if (error)
+    return NULL;
+
+  variant = g_variant_new_from_data (G_VARIANT_TYPE (AGGREGATE_TIMER_VARIANT_TYPE),
+                                     contents, size, FALSE, NULL, NULL);
+
+  if (!g_variant_is_normal_form (variant))
```
GeorgesStavracas, line 129: Following Will's comment, I'm assuming that a variant loaded from disk and not in normal form == file is corrupted. Is this a safe assumption to make here? Are there alternatives to checking if the file is corrupt?

daemon/emer-aggregate-timer-impl.c
```diff
@@ -0,0 +1,165 @@
...
+  g_autoptr(GVariant) cache_key = NULL;
+
+  self = g_object_new (EMER_TYPE_AGGREGATE_TIMER_IMPL, NULL);
+  self->tally = tally;
+  self->unix_user_id = unix_user_id;
+  self->event_id = g_variant_ref (event_id);
+  self->aggregate_key = g_variant_ref (aggregate_key);
+  self->payload = payload ? g_variant_ref (payload) : NULL;
+  self->start_monotonic_us = monotonic_time_us;
+  self->cache_key = g_variant_new ("(u@ayvmv)",
```
GeorgesStavracas, line 87: This variant is used to hash the aggregate timer, but it contains `mv` - is that a problem?

GeorgesStavracas (commented):
Commenting my own code so I don't lose my notes again. @pwithnall would you mind sharing your thoughts on these questions?

daemon/emer-aggregate-tally.c
wjt: If we are worried about this… can we reconsider using sqlite? :)

wjt: But what's the alternative (regardless of storage technology)? write a new file with each integer and sum them only when we come to submit the event? Append to the file (which on flash media rewrites the whole file)?

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,391 @@
...
+static GVariant*
+load_aggregate_timer_from_path (const char *path)
+{
+  g_autoptr(GVariant) variant = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree gchar *contents = NULL;
+  gsize size;
+
+  if (!g_file_test (path, G_FILE_TEST_EXISTS))
+    return NULL;
```
pwithnall, line 120: There’s a [TOCTTOU race](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) between this `g_file_test()` call and `g_file_get_contents()` below. Drop the `g_file_test()` call, and ignore the error from `g_file_get_contents()` if it’s `G_FILE_ERROR_NOENT`.

daemon/emer-aggregate-tally.c
pwithnall: Yes, that’s the right way to do things. Checking if a variant is in normal form will catch some forms of corruption, but can’t catch all corruption. If you want to do that, you’ll need to implement a checksum over the file. `is_normal_form()` basically checks whether some invariants of the variant packing format are correct (such as padding being correct and being all zero bytes).

daemon/emer-aggregate-tally.c
pwithnall: Is the file expected to be big (like, tens of kilobytes or larger)? From the looks of the code in this file, it isn’t expected to be big. I wouldn’t worry about it unless profiling shows that this is a performance problem in practice.

I guess any alternative, to be more performant, would have to load only the counter from disk. This code currently loads the entire file into a blob of heap-allocated memory, traverses over all of it checking it’s in normal form, and then indexes the counter from it. A more performant alternative would mmap the file and directly index the counter, or just store the counter in a file by itself.

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,391 @@
...
+  GObject parent_instance;
+
+  gchar *persistent_cache_directory;
+};
+
+G_DEFINE_TYPE (EmerAggregateTally, emer_aggregate_tally, G_TYPE_OBJECT)
+
+enum {
+  PROP_0,
+  PROP_PERSISTENT_CACHE_DIRECTORY,
```
pwithnall, line 40:
```suggestion
  PROP_PERSISTENT_CACHE_DIRECTORY = 1,
```

daemon/emer-aggregate-timer-impl.c
pwithnall: Yes, although the problem is actually that the variant is a container (tuple) at all — `g_variant_hash()` is defined only for basic variant types (i.e. not containers) and will emit a runtime warning when called on this.

You’ll need to build your own hash function, or choose a basic variant from within the cache key which can hash and represent the whole thing.

It’s fine to pass this type to `g_variant_equal()` though. In general, `maybe` types are fully supported by GVariant except for passing over D-Bus.

wjt: Do we have any cases where the payload is neither None nor the same as the aggregate key? I don't think so – [Daily|Monthly]AppUsage uses the same string, and time-in-session has None payload. So perhaps we can merge them in the D-Bus API as just the aggregate key + boolean for whether to send it as the payload to the server, and avoid the need to have a maybe here?

pwithnall: I might be misunderstanding what you’re suggesting, so to reiterate:
 - If you call `g_variant_hash()` on a variant of type `(u@ayvmv)` it will fail because the variant is a tuple, not because it contains a `maybe`
 - Calling `g_variant_hash()` on `(u@ayvsb)` or `(u@ayvv)` or `(*)` will fail similarly

wjt: I had misunderstood. Thanks!

tests/daemon/test-aggregate-tally.c
```diff
@@ -0,0 +1,194 @@
...
+
+  if (uuid_parse (event_id, parsed_event_id) != 0)
+    return NULL;
+
+  g_variant_builder_init (&event_id_builder, G_VARIANT_TYPE ("ay"));
+  for (size_t i = 0; i < UUID_LENGTH; ++i)
+    g_variant_builder_add (&event_id_builder, "y", parsed_event_id[i]);
+
+  return g_variant_new ("(aymv)", &event_id_builder, NULL);
+}
```
wjt, line 72: `get_uuid_builder` in metrics-util.h may be of interest

tests/daemon/test-aggregate-tally.c
```diff
@@ -0,0 +1,194 @@
...
+
+static void
+test_aggregate_tally_store_single_event (struct Fixture *fixture,
+                                         gconstpointer   dontuseme)
+{
+  struct AggregateEvent *event;
+
+  event = create_aggregate_event (0,
+                                  uuids[0],
+                                  g_variant_new ("v", g_variant_new_string ("")),
```
wjt, line 146: I don't think these need to be GVariants containing a single variant, they can be:

```suggestion
                                  g_variant_new_string ("")
```

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,391 @@
...
+}
+
+void
+emer_aggregate_tally_store_event (EmerAggregateTally *self,
+                                  guint32             unix_user_id,
+                                  GVariant           *event_id,
+                                  GVariant           *aggregate_key,
+                                  GVariant           *payload,
+                                  guint32             counter,
+                                  const char         *date,
```
wjt, line 253: I think it would be preferable to use `GDate *` in the API rather than this "stringly-typed" approach.

wjt (changes requested):

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,391 @@
...
+                                     const char         *date)
+{
+  g_autofree gchar *aggregate_key_str = NULL;
+  g_autofree gchar *event_id_str = NULL;
+  g_autofree gchar *filename = NULL;
+
+  aggregate_key_str = g_variant_print (aggregate_key, FALSE);
+  event_id_str = g_variant_print (event_id, FALSE);
+  filename = g_strdup_printf ("%u-%u-%u",
+                              g_str_hash (aggregate_key_str),
```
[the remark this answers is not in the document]
pwithnall: (I don’t know the full context surrounding this, but…)

In general it’s better (imo) to use escaping, rather than hashing, because that allows better debuggability. It’s scary to try and debug a directory full of files which are named after checksums. Less scary to debug a directory full of files with human-readable-but-escaped names.

daemon/emer-aggregate-tally.c
wjt: Oh… it's a string because it can be either yyyy-mm-dd or yyyy-mm. I still think we could stand to enforce one or the other in the API.

pwithnall: > I still think we could stand to enforce one or the other in the API.

Or define a helper type which enforces one or the other, and has some utility functions for converting to/from strings/`GDate`s/`GDateTime`s

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,391 @@
...
+                              unix_user_id,
+                              g_str_hash (event_id_str));
```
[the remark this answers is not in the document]
GeorgesStavracas: Hm, event ids are stored as `ay` in GVariants, and the documentation for this type mentions:

> The type of an array of bytes. This type is commonly used to pass around strings that may not be valid utf8.

I assume you're saying it's guaranteed because eos-metrics take a `const char*`?

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,391 @@
...
+  previous_counter =
+    find_previous_aggregate_timer_counter (self, unix_user_id,
+                                           event_id, aggregate_key,
+                                           date);
+
+  tally_path = get_tally_path_from_aggregate_timer (self, unix_user_id,
+                                                    event_id, aggregate_key,
+                                                    date);
+
+  ensure_folder_for_date (self, date, &error);
```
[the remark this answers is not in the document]
GeorgesStavracas: This, and there was another needless tally path build there too, which I removed.

daemon/emer-aggregate-timer-impl.c
GeorgesStavracas: I have worked around this by printing the cache key variant into a string, and comparing the strings. I'm assuming that two variants built equally should print the same string.

daemon/emer-aggregate-tally.c
GeorgesStavracas: Thanks

daemon/emer-aggregate-timer-impl.c
pwithnall: > I'm assuming that two variants built equally should print the same string.

Yes, if you are using `g_variant_print()`. That’s actually exactly how `g_variant_equal()` works!

daemon/emer-aggregate-tally.c
GeorgesStavracas: The file is expected to be pretty small, yeah.

daemon/emer-aggregate-tally.c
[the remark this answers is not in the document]
GeorgesStavracas: I've replaced it with a SHA256sum of all the fields we use to hash each timer.

daemon/emer-main.c
```diff
@@ -124,6 +124,79 @@ on_reset_tracking_id (EmerEventRecorderServer *server,
...
 
+static gboolean
+on_start_aggregate_timer (EmerEventRecorderServer *object,
+                          GDBusMethodInvocation   *invocation,
+                          GVariant                *event_id,
+                          GVariant                *aggregate_key,
+                          gboolean                 has_payload,
+                          GVariant                *payload,
+                          EmerDaemon              *daemon)
+{
```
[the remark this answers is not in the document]
wjt: This is still not addressed AFAICT.

tests/daemon/test-aggregate-tally.c
```diff
@@ -0,0 +1,253 @@
...
+    }
+}
+
+static void
+setup (struct Fixture *fixture,
+       gconstpointer   dontuseme)
+{
+  teardown (fixture, dontuseme);
+
+  g_mkdir (TEST_DIRECTORY, 0777);
```
wjt, line 117: It is not good practice to use a fixed, world-writable path for temporary directories, even in test code.

Happily you don't have to do this by hand. GLib has its own API for giving tests their own temporary directories which get cleaned up.

https://docs.gtk.org/glib/const.TEST_OPTION_ISOLATE_DIRS.html

The gist is that you pass it to `g_test_init`, and then you can simply use `g_get_xdg_cache_dir()` and friends and get a temporary directory that will be cleaned away.

tests/daemon/test-aggregate-tally.c
```diff
@@ -0,0 +1,253 @@
...
+
+static void
+test_aggregate_tally_iter (struct Fixture *fixture,
+                           gconstpointer   dontuseme)
+{
+  struct IterData data;
+  int i;
+
+  // Add the same aggregate event multiple times. It must
+  // result in a single aggregate event with the
```
wjt, line 197: This sentence is unfinished

daemon/emer-main.c
```diff
@@ -100,6 +100,91 @@ on_set_enabled (EmerEventRecorderServer *server,
...
 }
 
+static gboolean
+on_start_aggregate_timer (EmerEventRecorderServer *object,
+                          GDBusMethodInvocation   *invocation,
+                          GVariant                *event_id,
+                          GVariant                *aggregate_key,
+                          gboolean                 has_payload,
+                          GVariant                *payload,
+                          EmerDaemon              *daemon)
```
wjt, line 110: If an app calls this D-Bus method, and then disconnects from the bus without calling Stop (e.g. because it crashe), the timer will run forever.

The fix would be to use [g_bus_watch_name_on_connection](https://gnome.pages.gitlab.gnome.org/gtk/gio/func.bus_watch_name_on_connection.html) (connection, sender, G_BUS_NAME_WATCHER_FLAGS_NONE, NULL, vanished_cb, ...), stop the timer when the callback fires, and remove the watch when the timer is stopped.

This could be a follow-up branch but it does need to happen IMO.

wjt: I don't see a need for checking polkit connections for clients in this case.

wjt: I opened https://phabricator.endlessm.com/T32641 for this.

It is not a regression that gnome-shell crashing means running apps' time open so far is lost – that was true in the old stack. But now there is dangling state in the daemon, I think it's more important to clean up.

daemon/emer-main.c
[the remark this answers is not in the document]
GeorgesStavracas: Done

daemon/emer-main.c
```diff
@@ -100,6 +100,79 @@ on_set_enabled (EmerEventRecorderServer *server,
...
+  static guint64 timer_id = 0;
+  EmerAggregateTimer *timer;
+  const gchar *sender;
+  uid_t unix_user_id;
+
+  system_bus = g_dbus_method_invocation_get_connection (invocation);
+  timer_object_path = g_strdup_printf ("/com/endlessm/Metrics/AggregateTimer%lu",
+                                       timer_id);
+
+  timer = emer_aggregate_timer_skeleton_new ();
```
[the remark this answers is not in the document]
GeorgesStavracas: Should be fine now. Ownership goes to EmerAggregateTimerImpl (unfortunately seems like I can't use autoptr, so some manual unrefs were added to the code that runs before the corresponding EmerAggregateTimerImpl is created).

tests/daemon/test-aggregate-tally.c
```diff
@@ -0,0 +1,254 @@
...
+teardown (struct Fixture *fixture,
+          gconstpointer   dontuseme)
+{
+  g_clear_object (&fixture->tally);
+
+  if (fixture->tally_folder)
+    {
+      delete_folder (fixture->tally_folder, NULL, NULL);
+      g_clear_object (&fixture->tally_folder);
+    }
```
wjt, line 107: With G_TEST_OPTION_ISOLATE_DIRS this teardown code is not needed. GLib will delete the temporary directory for you. As a result `delete_folder` is also not needed.

wjt:
o glorious bot, retest this please now that https://github.com/endlessm/eos-metrics/pull/43 has landed

daemon/emer-daemon.c
```diff
@@ -1123,6 +1133,178 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+  EmerDaemonPrivate *priv = emer_daemon_get_instance_private (self);
+  g_autoptr(GDateTime) now = NULL;
+  g_autoptr(GError) error = NULL;
+  gint64 now_monotonic_us;
+
+  now = g_date_time_new_now_local ();
+  now_monotonic_us = g_get_monotonic_time ();
+  emer_aggregate_timer_impl_stop (timer_impl, now, now_monotonic_us, &error);
+  if (error)
+    g_warning ("Error stopping timer: %s", error->message);
```
wjt, line 1298: You could return this error to the caller

daemon/emer-daemon.c
```diff
@@ -1123,6 +1133,178 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+
+  emer_aggregate_timer_complete_stop_timer (timer, invocation);
+
+  if (!g_hash_table_remove (priv->aggregate_timers, timer_impl))
+    g_warning ("Failed to remove timer from set");
```
wjt, line 1303:
```suggestion
    g_warning ("Stopped timer was not in the set of running timers");
```

daemon/emer-main.c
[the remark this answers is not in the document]
wjt: Ah yes, an old friend. https://github.com/endlessm/eos-metrics/pull/44 should fix this.

daemon/emer-daemon.c
```diff
@@ -1665,3 +1866,74 @@ emer_daemon_get_permissions_provider (EmerDaemon *self)
...
+  g_object_set_data_full (G_OBJECT (timer_impl),
+                          "daemon",
+                          g_object_ref (self),
+                          g_object_unref);
+
+  if (g_hash_table_contains (priv->aggregate_timers, timer_impl))
+    {
+      g_warning ("There already is a timer for this event and user");
+      g_set_error (error, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+                   "There already is a timer for this event and user");
```
wjt, line 1927: event, user, aggregate_key and payload.

I don't think we need the warning. we are returning an error to the caller.

```suggestion
      g_set_error (error, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
                   "There already is a running timer for this set of arguments");
```

daemon/emer-daemon.c
```diff
@@ -1123,6 +1133,178 @@ set_max_bytes_buffered (EmerDaemon *self,
...
+      g_autoptr(GError) error = NULL;
+
+      emer_aggregate_timer_impl_store (timer_impl,
+                                       date,
+                                       monotonic_time_us,
+                                       &error);
+
+      /* Erroring here shouldn't stop the loop */
+      if (error)
+        g_warning ("Error splitting timer: %s", error->message);
```
wjt, line 1159:
```suggestion
        g_warning ("Error storing timer: %s", error->message);
```

daemon/emer-aggregate-timer-impl.c
```diff
@@ -0,0 +1,163 @@
...
+}
+
+void
+emer_aggregate_timer_impl_split (EmerAggregateTimerImpl *self,
+                                 gint64                  monotonic_time_us)
+{
+  g_return_if_fail (EMER_IS_AGGREGATE_TIMER_IMPL (self));
+
+  self->start_monotonic_us = monotonic_time_us;
+}
```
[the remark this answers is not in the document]
wjt: Got it!

daemon/emer-aggregate-tally.c
```diff
@@ -0,0 +1,393 @@
...
+                     (const guchar*) &unix_user_id,
+                     sizeof (guint32));
+  g_checksum_update (checksum,
+                     (const guchar*) aggregate_key_str,
+                     strlen (aggregate_key_str));
+
+  return g_build_filename (self->persistent_cache_directory,
+                           "aggregate-timers",
+                           date,
+                           g_checksum_get_string (checksum),
```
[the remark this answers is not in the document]
wjt: OK, let's leave this as it is

(For the record I agree that using the arbitrary aggregate key as part of a file name is a really bad idea. It's the other two which we know to be safe. But I won't die on this hill.)

wjt (commented):
OK, I now realise the build failure was due to https://github.com/endlessm/eos-metrics/pull/43 not being merged.

daemon/emer-daemon.c
GeorgesStavracas: Good idea

PR #258, opened 2021-10-13
Stop running timers

GeorgesStavracas:
WIP for now, but seems to be working.

https://phabricator.endlessm.com/T32641

### DIFF
--- a/daemon/emer-aggregate-timer-impl.c
+++ b/daemon/emer-aggregate-timer-impl.c
@@ -204,8 +204,7 @@ emer_aggregate_timer_impl_get_sender_name (EmerAggregateTimerImpl *self)
 guint
 emer_aggregate_timer_impl_hash (gconstpointer timer_impl)
 {
-  const EmerAggregateTimerImpl *self =
-    (EmerAggregateTimerImpl *)timer_impl;
+  const EmerAggregateTimerImpl *self = EMER_AGGREGATE_TIMER_IMPL ((gpointer)timer_impl);
 
   return g_str_hash (self->cache_key_string);
 }
@@ -214,8 +213,8 @@ gboolean
 emer_aggregate_timer_impl_equal (gconstpointer a,
                                  gconstpointer b)
 {
-  const EmerAggregateTimerImpl *timer_impl_a = (EmerAggregateTimerImpl *)a;
-  const EmerAggregateTimerImpl *timer_impl_b = (EmerAggregateTimerImpl *)b;
+  const EmerAggregateTimerImpl *timer_impl_a = EMER_AGGREGATE_TIMER_IMPL ((gpointer)a);
+  const EmerAggregateTimerImpl *timer_impl_b = EMER_AGGREGATE_TIMER_IMPL ((gpointer)b);
 
   return g_str_equal (timer_impl_a->cache_key_string,
                       timer_impl_b->cache_key_string);


daemon/emer-daemon.c
```diff
@@ -1933,5 +2010,27 @@ emer_daemon_start_aggregate_timer (EmerDaemon          *self,
 
   g_hash_table_add (priv->aggregate_timers, g_steal_pointer (&timer_impl));
 
+  /* Monitor sender and shut down all timers from it when it vanishes */
+  sender_data = g_hash_table_lookup (priv->monitored_senders, sender_name);
```
wjt, line 2014: Drive-by thought: does GDBus do the same thing internally, in which case there is no advantage to coalescing senders together here?

GeorgesStavracas: Looking at https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gdbusnamewatching.c, I don't see this happening. But, I think it's actually good to do that ourselves - it's easy to do, literally involves 4 lines of code.

GeorgesStavracas:
This is ready for review now. I've tested this by killing GNOME Shell (testing if the vanish callback works), and by opening various apps then closing them all (testing if the D-Bus watch is removed).

daemon/emer-daemon.c
```diff
@@ -1446,6 +1515,7 @@ emer_daemon_finalize (GObject *object)
   g_warn_if_fail (priv->current_upload_cancellable == NULL);
 
   g_clear_pointer (&priv->aggregate_timers, g_hash_table_destroy);
+  g_clear_pointer (&priv->monitored_senders, g_hash_table_destroy);
```
wjt, line 1518: Is this the right order to clean these up?

After clearing `aggregate_timers`, `monitored_senders`'s values will contain GPtrArrays of `EmerAggregateTimerImpl`. But those arrays do not own a reference to those, so they are likely dangling pointers.

OTOH the cleanup function doesn't access them.

daemon/emer-daemon.c
```diff
@@ -1299,12 +1320,60 @@ on_timer_stopped_cb (EmerAggregateTimer     *timer,
...
+
+      if (error)
+        g_warning ("Error stopping aggregate timer after sender disappeared: %s",
+                   error->message);
+
+      if (!g_hash_table_remove (priv->aggregate_timers, timer_impl))
+        g_warning ("Stopped timer was not in the set of running timers");
+    }
+
+  g_hash_table_remove (priv->aggregate_timers, sender_name);
```
wjt, line 1374: This is the wrong hash table. `aggregate_timers` is a map-as-set whose keys are EmerAggregateTimerImpl.

I am surprised this doesn't crash when calling emer_aggregate_timer_impl_hash on a string!

Before fixing this bug, I suggest adjusting `emer_aggregate_timer_impl_hash` and `emer_aggregate_timer_impl_equal` to use checked casts of their arguments:

```diff
diff --git a/daemon/emer-aggregate-timer-impl.c b/daemon/emer-aggregate-timer-impl.c
index 4294a09..5ebeede 100644
--- a/daemon/emer-aggregate-timer-impl.c
+++ b/daemon/emer-aggregate-timer-impl.c
@@ -192,8 +192,7 @@ emer_aggregate_timer_impl_stop (EmerAggregateTimerImpl  *self,
 guint
 emer_aggregate_timer_impl_hash (gconstpointer timer_impl)
 {
-  const EmerAggregateTimerImpl *self =
-    (EmerAggregateTimerImpl *)timer_impl;
+  const EmerAggregateTimerImpl *self = EMER_AGGREGATE_TIMER_IMPL (timer_impl);
 
   return g_str_hash (self->cache_key_string);
 }
@@ -202,8 +201,8 @@ gboolean
 emer_aggregate_timer_impl_equal (gconstpointer a,
                                  gconstpointer b)
 {
-  const EmerAggregateTimerImpl *timer_impl_a = (EmerAggregateTimerImpl *)a;
-  const EmerAggregateTimerImpl *timer_impl_b = (EmerAggregateTimerImpl *)b;
+  const EmerAggregateTimerImpl *timer_impl_a = EMER_AGGREGATE_TIMER_IMPL (a);
+  const EmerAggregateTimerImpl *timer_impl_b = EMER_AGGREGATE_TIMER_IMPL (b);
 
   return g_str_equal (timer_impl_a->cache_key_string,
                       timer_impl_b->cache_key_string);
```

Then testing the code should log a critical.

wjt:
Please add a test.

One way you could test this is to create a new GDBusConnection to the bus, pass its unique name into the function, then close it, and await the daemon stopping its timer.

PR #261, opened 2021-10-21
aggregate-timer-impl: Convert counter to seconds

GeorgesStavracas:
EmerAggregateTimerImpl tracks the amount of passed time
using the monotonic clock, which gives us a time counter
in μs. However, our plumbing expects the counter to be
in seconds.

Convert the ellapsed aggregate time from μs to seconds
before storing and sending it.

https://phabricator.endlessm.com/T32203

### DIFF
--- a/daemon/emer-aggregate-timer-impl.c
+++ b/daemon/emer-aggregate-timer-impl.c
@@ -112,12 +112,15 @@ emer_aggregate_timer_impl_store (EmerAggregateTimerImpl  *self,
                                  gint64                   monotonic_time_us,
                                  GError                 **error)
 {
-  gint64 counter;
+  guint32 counter;
+  gint64 difference;
 
   g_return_val_if_fail (EMER_IS_AGGREGATE_TIMER_IMPL (self), FALSE);
   g_return_val_if_fail (date != NULL && *date != '\0', FALSE);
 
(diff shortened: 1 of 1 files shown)


daemon/emer-aggregate-timer-impl.c
```diff
@@ -112,12 +112,15 @@ emer_aggregate_timer_impl_store (EmerAggregateTimerImpl  *self,
...
+  guint32 counter;
+  gint64 difference;
 
   g_return_val_if_fail (EMER_IS_AGGREGATE_TIMER_IMPL (self), FALSE);
   g_return_val_if_fail (date != NULL && *date != '\0', FALSE);
 
-  counter = monotonic_time_us - self->start_monotonic_us;
+  difference = monotonic_time_us - self->start_monotonic_us;
+  counter = (guint32) (difference / G_USEC_PER_SEC);
+
```
wjt, line 123: Perhaps `CLAMP (that sum, 0, G_MAXUINT32)`?

wjt: Rather than the cast

wjt (commented):
Looks fine, I think CLAMP would be an improvement -- ok to merge after that change.

PR #263, opened 2021-10-26
SQLite

GeorgesStavracas:
Port the tally to SQLite, and add an extra test case to make sure we don't regress on deleting tallies.

https://phabricator.endlessm.com/T32661


GeorgesStavracas:
The super annoying issue that I've been fighting for the past few days turned out to be BLOB vs TEXT. When inserting data into the database, Will's patch was passing the printed GVariant string to what SQLite treats as BLOBs, which was stored just fine, but would fail to match any rows when doing a SELECT or a DELETE.

daemon/emer-aggregate-tally.c
```diff
@@ -92,97 +103,131 @@ aggregate_timer_to_variant (guint32    unix_user_id,
...
 
-  variant = load_aggregate_timer_from_path (aggregate_counter_path);
+  G_OBJECT_CLASS (emer_aggregate_tally_parent_class)->constructed (object);
 
-  if (variant)
-    g_variant_get_child (variant, 3, "u", &previous_counter);
+  g_assert (self->persistent_cache_directory != NULL);
+  ensure_folder_exists (self, self->persistent_cache_directory, /* TODO */ NULL);
+  path = g_build_filename (self->persistent_cache_directory,
+                           "metrics.db",
```
dbnicholson, line 191: I was going to suggest that this be named `aggregate-events.db` or something since that's all it's used for, but I could see this gaining wider usage later.

dbnicholson (commented):
I think this looks to be getting to a good place. I'm neither a SQL nor sqlite guru, but I had a few questions around that part.

daemon/emer-aggregate-tally.c
```diff
@@ -92,97 +103,131 @@ aggregate_timer_to_variant (guint32    unix_user_id,
...
+   * https://sqlite.org/wal.html
+   */
+  tally_exec_or_die (self, "PRAGMA journal_mode = WAL");
+  /* Magic number is "emer" in ASCII */
+  tally_exec_or_die (self, "PRAGMA application_id = 0x656d6572");
+  tally_exec_or_die (self, "CREATE TABLE IF NOT EXISTS tally (date TEXT NOT NULL,"
+                           "                                  event_id TEXT NOT NULL,"
+                           "                                  unix_user_id INT NOT NULL,"
+                           "                                  aggregate_key TEXT NOT NULL,"
+                           "                                  payload TEXT,"
```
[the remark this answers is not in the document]
wjt: Unknown events are still stored in this way in Azafea

daemon/emer-aggregate-tally.c
```diff
@@ -92,97 +103,131 @@ aggregate_timer_to_variant (guint32    unix_user_id,
...
+   *
+   * TODO: should we force a checkpoint/fsync() to disk periodically, e.g.
+   * whenever we try to submit to the server?
+   *
+   * https://sqlite.org/wal.html
+   */
+  tally_exec_or_die (self, "PRAGMA journal_mode = WAL");
+  /* Magic number is "emer" in ASCII */
+  tally_exec_or_die (self, "PRAGMA application_id = 0x656d6572");
+  tally_exec_or_die (self, "CREATE TABLE IF NOT EXISTS tally (date TEXT NOT NULL,"
```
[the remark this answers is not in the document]
GeorgesStavracas: If I add a primary key to each row, I'll also have to do a SELECT before each INSERT

daemon/emer-aggregate-tally.c
```diff
@@ -259,50 +277,29 @@ emer_aggregate_tally_store_event (EmerAggregateTally  *self,
...
+  CHECK (sqlite3_prepare_v2 (self->db, UPSERT_SQL, -1, &stmt, NULL));
 
-  timer_variant = aggregate_timer_to_variant (unix_user_id, event_id,
-                                              aggregate_key, payload,
-                                              counter + previous_counter,
-                                              monotonic_time_us);
+  CHECK (sqlite3_bind_text (stmt, 1, date, -1, SQLITE_TRANSIENT));
+  CHECK (sqlite3_bind_text (stmt, 2, g_variant_print (event_id, TRUE), -1, g_free));
+  CHECK (sqlite3_bind_int (stmt, 3, * ((int *) &unix_user_id)));
+  CHECK (sqlite3_bind_blob (stmt, 4, g_variant_print (aggregate_key, TRUE), -1, g_free));
```
dbnicholson, line 296: Why are you storing the printed variant as a blob? Why not `g_variant_get_data()`?

daemon/emer-aggregate-tally.c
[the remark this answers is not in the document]
wjt: Why would you have to do that?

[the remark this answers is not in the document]
wjt: You don't need to query for the largest previous number – you can declare the column to be an autoincrementing integer, and don't specify it in the `INSERT`.

rowid, a weird sqlite-ism, is an implicit autoincrementing primary key that every table without an explicit primary key has.

daemon/emer-aggregate-tally.c
```diff
@@ -76,113 +86,132 @@ ensure_folder_exists (EmerAggregateTally  *self,
...
 static guint32
-find_previous_aggregate_timer_counter (EmerAggregateTally *self,
-                                       const char         *aggregate_counter_path)
+column_to_uint32 (sqlite3_stmt *stmt,
+                  int           i)
 {
-  g_autoptr(GVariant) variant = NULL;
-  guint32 previous_counter = 0;
+  int number = sqlite3_column_int (stmt, i);
+  return *((guint32*) &number);
```
wjt, line 153: `CLAMP (number, 0, MAXUINT32)`

daemon/emer-aggregate-tally.c
```diff
@@ -315,101 +340,74 @@ emer_aggregate_tally_iter (EmerAggregateTally *self,
...
         break;
     }
 
-  if (flags & EMER_TALLY_ITER_FLAG_DELETE)
+  CHECK (sqlite3_finalize (stmt));
+
+  if (rows_to_delete->len > 0)
     {
-      g_autoptr(GError) deletion_error = NULL;
+      const char *DELETE_SQL = "DELETE FROM tally WHERE id = ?;";
```
wjt, line 395: An optimization that I have seen done is to generate a query of the form:

```
DELETE FROM tally WHERE id = (?, ?, ?, ...);
```

i.e. the number of `?` is dynamically generated. Then you bind all the IDs and execute a single statement. But this is ugly to implement if the sql binding doesn't provide it, (I don't think sqlite does,) and in this case I think doing it one at a time is fine.

daemon/emer-aggregate-tally.c
```diff
@@ -76,113 +86,132 @@ ensure_folder_exists (EmerAggregateTally  *self,
...
+                           "                                  aggregate_key_type TEXT NOT NULL,"
+                           "                                  aggregate_key BLOB NOT NULL,"
+                           "                                  payload_type TEXT,"
+                           "                                  payload BLOB,"
+                           "                                  counter INT NOT NULL);");
+  tally_exec_or_die (self,
+                     "CREATE UNIQUE INDEX IF NOT EXISTS "
+                     "ix_tally_unique_fields ON tally (date, event_id, unix_user_id, "
+                     "                                 aggregate_key_type, aggregate_key, "
+                     "                                 payload_type, payload)");
```
wjt, line 201: I should have added this when I wrote the draft but I think we should set:

```
  tally_exec_or_die (self, "PRAGMA user_version = 1");
```

If we never use it, it's harmless, but if we later want to use it to do versioned schema migrations it may be helpful to have.

daemon/emer-aggregate-tally.c
```diff
@@ -259,50 +289,45 @@ emer_aggregate_tally_store_event (EmerAggregateTally  *self,
...
-    {
-      g_propagate_error (error, g_steal_pointer (&local_error));
-      g_prefix_error (error, "Error saving aggregate timer to tally: ");
-      return FALSE;
-    }
+  CHECK (sqlite3_prepare_v2 (self->db, UPSERT_SQL, -1, &stmt, NULL));
+
+  CHECK (sqlite3_bind_text (stmt, 1, date, -1, SQLITE_TRANSIENT));
+  CHECK (sqlite3_bind_text (stmt, 2, g_variant_print (event_id, TRUE), -1, g_free));
+  CHECK (sqlite3_bind_int (stmt, 3, * ((int *) &unix_user_id)));
```
wjt, line 311: Why the cast via a pointer? Why not:

```suggestion
  CHECK (sqlite3_bind_int (stmt, 3, (int) unix_user_id));
```

?

wjt (commented):
Apart from the unusual casts between int and uint32, this looks great

daemon/emer-aggregate-tally.c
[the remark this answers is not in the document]
wjt: Ah, I see you solved this :)

daemon/emer-aggregate-tally.c
```diff
@@ -259,50 +289,45 @@ emer_aggregate_tally_store_event (EmerAggregateTally  *self,
...
+  CHECK (sqlite3_bind_text (stmt, 4,
+                            g_variant_get_type_string (aggregate_key),
+                            -1, SQLITE_TRANSIENT));
+  CHECK (sqlite3_bind_blob (stmt, 5,
+                            g_variant_get_data (aggregate_key),
+                            g_variant_get_size (aggregate_key),
+                            SQLITE_TRANSIENT));
```
wjt, line 318: I realised that an alternative to explicitly storing the type and data separately would be to store:

```g_variant_get_data (g_variant_new_variant (aggregate_key))```

since GVariants of type `v` carry the type of their contents. Then on the way out, load a variant of type `v`, then get the child element of the variant. But that's functionally equivalent to what you have here – and breaking out the type string into a column makes the db easier to inspect interactively.

daemon/emer-aggregate-tally.c
GeorgesStavracas: This simple cast loses 1 bit to the signal, what I'm doing here is trick C into copying all bits of uint32 into an integer - even if the first bit is `1`, which an integer would interpret as a negative number.

daemon/emer-aggregate-tally.c
GeorgesStavracas: This is not necessary (and will make the compiler complain about signedness comparison) because the type size doesn't allow anything smaller than 0, nor bigger than, well, the maximum number 1s it can fit.

GeorgesStavracas: I've made this a bit more readable.

daemon/emer-aggregate-tally.c
GeorgesStavracas: Could you be more clear on whether you want me to implement it this way or not?

wjt: I think the way you have implemented it here is fine.

daemon/emer-aggregate-tally.c
wjt: Oh, now I understand!

However there is an easier way to achieve this. sqlite's integer type has the range of a 64-bit signed integer. https://www.sqlite.org/datatype3.html

There are `_int64` functions for working with this. I will push a patch which makes this change, adds a test for the case of adding 1 to MAXINT32, and also uses 64-bit ints for rowids.

wjt:
This looks good to me now, please feel free to merge if you approve of my additional patches!

GeorgesStavracas:
GitHub now forbids authors to approve their own pull requests, but the last two patches are approved.

PR #266, opened 2021-11-19
Refcount timers

GeorgesStavracas:
Quoting the last commit

```
Instead of failing straight away when there's a running
timer, increase its run-count; and decrease it whenever
StopTimer() is called. When the run count reaches zero,
that's when the timer object is unexported and we save
the aggregated timer into the tally.
```

https://phabricator.endlessm.com/T32671

### DIFF
--- a/daemon/emer-aggregate-timer-impl.c
+++ b/daemon/emer-aggregate-timer-impl.c
@@ -114,7 +114,8 @@ emer_aggregate_timer_impl_new (EmerAggregateTally *tally,
   if (payload)
     g_variant_take_ref (payload);
 
-  cache_key = g_variant_new ("(u@ayvmv)",
+  cache_key = g_variant_new ("(su@ayvmv)",
+                             sender_name,
                              unix_user_id,
                              event_id,
                              aggregate_key,
--- a/daemon/emer-aggregate-timer-impl.c
+++ b/daemon/emer-aggregate-timer-impl.c
@@ -42,6 +42,8 @@ struct _EmerAggregateTimerImpl
   GVariant *payload; /* owned */
   gchar *cache_key_string; /* owned */
   gchar *sender_name; /* owned */
+
+  guint32 run_count;
 };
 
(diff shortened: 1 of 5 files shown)


daemon/emer-aggregate-timer-impl.c
```diff
@@ -114,7 +114,8 @@ emer_aggregate_timer_impl_new (EmerAggregateTally *tally,
   if (payload)
     g_variant_take_ref (payload);
 
-  cache_key = g_variant_new ("(u@ayvmv)",
+  cache_key = g_variant_new ("(su@ayvmv)",
+                             sender_name,
                              unix_user_id,
```
wjt: This is fine, but actually you don't need both the sender_name and unix_user_id in this key, since there is exactly 1 Unix user ID for a D-Bus unique name. (He says, shortly before wondering what happens if you connect to the bus as root, fork, and change your effective user id...)

daemon/emer-daemon.c
```diff
@@ -2050,7 +2051,14 @@ emer_daemon_start_aggregate_timer (EmerDaemon       *self,
...
 
-  if (g_hash_table_contains (priv->aggregate_timers, timer_impl))
+  timer_hash_string =
+    emer_aggregate_timer_impl_compose_hash_string (sender_name,
+                                                   unix_user_id,
+                                                   event_id,
+                                                   aggregate_key,
+                                                   nullable_payload);
+
+  if (g_hash_table_contains (priv->aggregate_timers, timer_hash_string))
```
wjt: The error message below can now be adjusted to include this hash string (which also makes it more accurate since the sender_name is now part of the unique-ifying key)

wjt (approved):
LGTM. I'll push a fixup! for my suggested change

daemon/emer-daemon.c
wjt: Oh wait, of course, this message is no longer present. Nice.

PR #268, opened 2022-01-13
print-persistent-cache: Show message on success

wjt:
I was baffled because it did not, in fact, print the persistent cache as
the name suggested. In fact it saves it to a file.


dbnicholson:
retest this please

dbnicholson (approved):
This looks entirely reasonable. I kicked the CI again to make sure it doesn't introduce any failures, though. Any reason not to merge it?

wjt:
Good to merge from my perspective. I don't think this tool is really used… in fact the test suite doesn't run it so we will see its coverage is 0% :)

PR #270, opened 2022-01-13
Define GLib API level macros

wjt:
Endless OS master now has GLib 2.70. Endless OS 4 has 2.66. Some
functions were deprecated between those two versions:

- g_memdup() is deprecated in favour of g_memdup2()
- g_binding_get_source() is deprecated in favour of
  g_binding_dup_source()

This project builds with fatal warnings by default. As a result this
project no longer builds on Endless OS master.

Because we still have outstanding work to land on the eos4 branch,
rather than fixing these warnings and bumping the minimum GLib version,
instead use GLIB_VERSION_MAX_ALLOWED to suppress these warnings (and
prevent inadvertantly using newer API). In future we can bump this and
fix the warnings.

https://phabricator.endlessm.com/T32967


dbnicholson (approved):
Seems reasonable.

PR #271, opened 2022-01-14
configure: Don't require --with-default-metrics-server

wjt:
2fa2c669ed5e313491f23579ea8a9af0791a0d48 (in
https://github.com/endlessm/eos-event-recorder-daemon/pull/249)
introduced the ability to configure the metrics server URL at build
time, and removed the default being to submit to the Endless server, to
ensure that development builds and builds for other distros don't
accidentally send metrics to Endless.

At the time, --with-default-metrics-server was made mandatory, with the
argument that people don't look at warnings from configure scripts.
While true, I think the other downsides outweigh this. It is desirable
that a project uses a standard build system and can be successfully
built and tested out of the box. Every time I clone this repo, I am
annoyed that I have to set this flag, and have to think about what value
to use during development. We had to hardcode a dummy value in the build
system for 'make distcheck' to pass, and another value in our Jenkins
configuration to allow it to configure the project in the first place in
order to run 'make distcheck'.

Now that I find myself wanting to switch this project to Meson, the
Jenkins stuff is an active impediment, because Meson doesn't accept the
same configuration arguments, so I would need to update the Jenkins
config in lockstep and find some way to maintain build support for older
OS branches.

Instead, let's define an obviously-dummy value as the default, the same
one as used by 'make distcheck', and emit a warning. 'azafea.example'
will never resolve and is obviously a placeholder, so you'll notice soon
enough in the journal that no metrics are being submitted (and they'll
be buffered on disk).

Includes patch from #270 to fix the build.

https://phabricator.endlessm.com/T32969

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -41,11 +41,10 @@ AM_CFLAGS = @WARN_CFLAGS@
 AM_LDFLAGS = @WARN_LDFLAGS@
 
 # Enable strict compiler flags and install systemd unit files to the specified
-# directory when doing 'make distcheck'. Set a dummy default server URL.
+# directory when doing 'make distcheck'.
 AM_DISTCHECK_CONFIGURE_FLAGS = \
 	--enable-compile-warnings=error \
 	--with-systemdsystemunitdir='$${prefix}/lib/systemd/system' \
-	--with-default-metrics-server=azafea.example \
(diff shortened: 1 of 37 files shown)


dbnicholson (approved):
Agreed.

PR #273, opened 2022-02-11
Switch to Meson build system

wjt:
https://phabricator.endlessm.com/T32969

Packaging counterpart on #274


eli-schwartz:
> Because Meson requires out-of-tree builds, developers can just put their favourite name for that directory in their global gitignore configuration (~/.config/git/ignore) and never think about this again.

Even better, Meson generates a `.gitignore` file inside the builddir, which ignores `*` for you.

It's something we know we can do, so why not do it? :)

wjt:
> Even better, Meson generates a `.gitignore` file inside the builddir, which ignores `*` for you.
> 
> It's something we know we can do, so why not do it? :)

I never knew that! It's such a good idea that I taught flatpak to do the same thing :) https://github.com/flatpak/flatpak/pull/4741

eli-schwartz:
Cool! Happy to be of inspiration.

Incidentally Meson also treats mercurial the same way. This doesn't matter for eos-event-recorder-daemon but maybe flatpak should create a .hgignore too?

Makefile.am
```diff
@@ -1,151 +0,0 @@
...
-			echo "*************************************"; \
-		fi; \
-	else \
-		echo "*******************************************"; \
-		echo "There is no 'metrics' user on your system."; \
-		echo "If this is an EndlessOS system, upgrade the"; \
-		echo "'base-passwd' package."; \
-		echo "Otherwise, create a 'metrics' user."; \
-		echo "*******************************************"; \
-	fi
```
wjt, line 115: Did not bother to try to preserve this since it is done by the tmpfiles snippet.

eli-schwartz: I assume those directories simply don't need to exist at all in the `ninja install` route?

e.g. if created in the Makefile, you'd end up with distro packages actually listing an empty directory here in the package manifest, and then uninstallation might try to remove it.

Meson can do this too, if you really want it, via install_emptydir (you'd need to bump your minimum meson from 0.52.0 to 0.60.0 though).

wjt: They don't need to exist. They are created at boot time by `systemd-tmpfiles`. (Actually on Endless OS any files placed in `/var` by packages at install time are completely ignored, so the empty directories are purely a waste of inodes.)

meson.build
```diff
@@ -0,0 +1,73 @@
...
+)
+
+systemd_dep = dependency('systemd')
+event_recorder_server_xml = eosmetrics_dep.get_pkgconfig_variable('datadir') / 'eos-metrics' / 'com.endlessm.Metrics.xml'
+
+default_metrics_server = get_option('default_metrics_server')
+
+prefix = get_option('prefix')
+libexec_dir = join_paths(prefix, get_option('libexecdir'))
+# TODO: remove trailing slash, inherited from Autotools build
```
pwithnall, line 45: Is this `TODO` comment something you want to address before this MR is merged?

meson.build
```diff
@@ -0,0 +1,73 @@
...
+emer_shared_required_modules = [
+    dependency('gio-unix-2.0', version: '>= 2.46'),
+    dependency('gobject-2.0', version: '>= 2.58'),
+    dependency('sqlite3'),
+    dependency('uuid'),
+]
+emer_required_modules = emer_shared_required_modules + [
+    eosmetrics_dep,
+    cc.find_library('m', required: false),
+    dependency('glib-2.0', version: '>= 2.63.1'),
```
pwithnall, line 23: Being pedantic, `glib-2.0` is almost certainly a shared required module, not a private one

configure.ac
```diff
@@ -1,189 +0,0 @@
...
-
-# Configure options
-# -----------------
-# --enable-compile-warnings. By default, warnings are always fatal. Documentation:
-# https://www.gnu.org/software/autoconf-archive/ax_compiler_flags.html
-#
-# -Wno-declaration-after-statement:
-#   Mixed decls are used a lot and are harmless.
-AX_COMPILER_FLAGS([WARN_CFLAGS], [WARN_LDFLAGS], [no], [],
-                  [-Wno-declaration-after-statement])
```
pwithnall, line 121: I guess dropping this option is covered by enabling `c_std=c11` in the Meson project options?

You may want to consider also setting `warning_level=3` (or some other value) in Meson’s `project()`.

data/meson.build
```diff
@@ -0,0 +1,52 @@
...
+    }),
+    install: true,
+    install_dir: systemd_dep.get_pkgconfig_variable('tmpfilesdir'),
+)
+
+# Daemon configuration
+install_data(
+    'eos-metrics-permissions.conf',
+    install_dir: config_dir,
+    install_mode: ['rwxrwsr-x', 'metrics', 'metrics'],
```
pwithnall, line 45: Given the config file is world-readable, does it have to be chowned to `metrics`? That seems a bit unusual. Every single file in `/etc` on my system is owned by `root`.

This is how the autotools system did things, though, so this is more of a passing comment/query than a request to change things.

tests/meson.build
```diff
@@ -0,0 +1,97 @@
...
+    args: [test_daemon],
+    # TODO: These tests are slow because they set the daemon's upload interval
+    # to 2 seconds and constantly wait for this time to elapse, rather than
+    # mocking the clock or telling the daemon to upload. Sure, *some* of them
+    # should exercise the daemon's internal timer, but not all.
+    timeout: 60,
+)
+
+
+# TODO: Logs loads of warnings, does this do the same thing in the Autotools build?
```
pwithnall, line 91: Looks like this TODO comment should be addressed before merging?

pwithnall (changes requested):
* “.gitignore: Remove!” — the commit message needs to be updated cf the discussion about Meson doing this automatically.

meson.build
wjt: Nope, I was going for zero source changes.

pwithnall: That makes sense!

data/meson.build
pwithnall: > the daemon modifies its own configuration in some situations

:scream:

That somehow feels wrong. But also definitely not something to address in this MR (and it might actually be the best approach anyway; I haven’t thought about it at all, and am just being reactionary).

configure.ac
wjt: Yes, I think so. Certainly I don't get warned about mixed declarations.

- `warning_level=3` enables `-Wpedantic` which is upset about some indeed pedantic stuff which I would rather not fix today.
- `warning_level=2` enables `-Wextra`  which spews warnings about unused parameters; adding `-Wno-unused-parameter` makes the build clean. So I'll set this option.

It doesn't seem straightforward to preserve the "fatal warnings for git checkout, non-fatal otherwise" default.

meson.build
wjt: It's not actively harmful, it's just not to my taste :)

data/meson.build
wjt: I had the same reaction!

tests/meson.build
wjt: Good catch!

pwithnall (approved):

PR #276, opened 2022-03-07
Adjust StartAggregateTimer, adding UID and removing aggregate key

wjt:
Depends on https://github.com/endlessm/eos-metrics/pull/47.

https://phabricator.endlessm.com/T33252


dbnicholson (commented):
Looks good. Continuing on with other reviews.

dbnicholson (approved):

PR #277, opened 2022-03-14
Set G_DEBUG=fatal-warnings in tests

wjt:
This requires two bits of cleanup to remove some warnings routinely emitted in one of the tests.


wjt:
This will have the side-effect of causing a rebuild which will fix a problem caused #276 being built against a too-old version of eos-metrics. The fact that it compiled and passed its test is a gap in test coverage that I plan to solve in due course :-)

pwithnall (approved):

PR #278, opened 2022-03-16
Smoke-test StartAggregateTimer over D-Bus

wjt:
This just crudely tests two easy cases of this method, and does so over D-Bus for simplicity.

It includes some changes to the error paths of that function, and also cleans up some dead code I missed in #276.

(There are more changes needed to discard running timers, and chuck away cached data, when metrics are disabled.)


dbnicholson (approved):
Totally worthy. At least it's going to get most of the code to run, which is better than testing it exclusively manually.

PR #279, opened 2022-03-19
Clear tally when metrics are disabled

wjt:
Previously, disabling metrics would cause any running aggregate timers
to be stopped, and any aggregate events from previous days/months which
are already in the EmerPersistentCache to be discarded, but events from
the current day/month would remain in the EmerAggregateTally.
Re-enabling metrics would cause these to be sent in due course.

Disabling metrics should remove all buffered data. Clear the tally too.

This also adds a test that validates the first point above.


wjt:
retest this please

dbnicholson (approved):
Nice tests!

PR #280, opened 2022-03-24
Don't critical on aggregates with NULL payload

wjt:
Previously, when aggregate events from previous days & months are taken from the tally and enqueued for submission (which occurs at midnight or when the computer starts up), if any of those events has a NULL payload, the daemon would log a pair of CRITICAL messages due to attempting to unref a NULL `GVariant *`.

The underlying cause is the code to fake the 'maybe' type over D-Bus with a (boolean, variant) pair. But the offending function is no longer exposed over D-Bus, so the signature can be simplified.

While I was here I finally understood why event payloads reach azafea with two levels of variant wrapper rather than one, and fixed it.

I also added a minimal test for enqueueing events from the tally on startup. I have not added a test for enqueueing events at midnight because this would require a more invasive refactoring to be able to mock the passage of time in tests.

https://phabricator.endlessm.com/T33299

### DIFF
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -800,6 +800,8 @@ static GVariant *
 get_nullable_payload (GVariant *payload,
                       gboolean  has_payload)
 {
+  g_return_val_if_fail (payload != NULL, NULL);
+
   if (!has_payload)
     {
       g_variant_ref_sink (payload);
@@ -1219,12 +1221,11 @@ buffer_aggregate_event_to_queue (guint32     unix_user_id,
 {
   EmerDaemon *self = user_data;
 
-  emer_daemon_record_aggregate_event (self,
-                                      get_uuid_as_variant (event_uuid),
-                                      date,
-                                      counter,
-                                      payload != NULL,
-                                      payload);
+  emer_daemon_enqueue_aggregate_event (self,
+                                       get_uuid_as_variant (event_uuid),
+                                       date,
+                                       counter,
+                                       payload);
 
   return EMER_TALLY_ITER_CONTINUE;
 }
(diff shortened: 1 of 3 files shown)


daemon/emer-daemon.c
```diff
@@ -1726,6 +1746,21 @@ emer_daemon_class_init (EmerDaemonClass *klass)
                          G_PARAM_CONSTRUCT_ONLY | G_PARAM_READWRITE |
                          G_PARAM_STATIC_STRINGS);
 
+  /*
+   * EmerDaemon:aggregate-tally:
```
pwithnall, line 1751:
```suggestion
   * EmerDaemon:aggregate-tally: (nullable)
```

pwithnall (approved):

daemon/emer-daemon.c
wjt: It can be `NULL` at construct time, but will always be non-`NULL` on a constructed object. Is this the right annotation for that case? (Maybe that cannot be expressed.)

wjt: https://gi.readthedocs.io/en/latest/annotations/giannotations.html#type-signature suggests there is no way to express this.

wjt:
Thanks!

daemon/emer-daemon.c
pwithnall: errrrrr, I actually meant to type `(not nullable)`, but got distracted by having to write out the GitHub suggestion syntax manually. Sorry.

`(not nullable)` is a valid annotation.

pwithnall: No, I was right. I did intend to type `(nullable)` because you can pass `NULL` in at construction time.

PR #281, opened 2022-03-30
daemon: Don't crash at startup if metrics are disabled

wjt:
Since 898c773 ("daemon: Clear tally when metrics are disabled"),
disabling metrics would cause the daemon to crash on startup. Here's the
chain of events:

1. set_permissions_provider(), called at construct time, connects to the
   notify::daemon-enabled signal on the EmerPermissionsProvider.
2. In emer_daemon_constructed(), if no EmerAggregateTally was provided
   as a construct-time property, priv->aggregate_tally was NULL until
   after a call to emer_permissions_provider_get_environment().
3. emer_permissions_provider_get_environment() causes the
   EmerPermissionsProvider to re-read its configuration file, and to
   unconditionally emit notify::daemon-enabled and
   notify::uploading-enabled even if those properties have not really
   changed.
4. on_permissions_changed() updates priv->recording_enabled and, if it
   is now FALSE, calls emer_aggregate_tally_clear(priv->aggregate_tally).
5. emer_aggregate_tally_clear() calls
      g_return_val_if_fail (EMER_IS_AGGREGATE_TALLY (self), FALSE);
   so returns FALSE with *error unset if called on a NULL pointer.
6. If emer_aggregate_tally_clear() fails, on_permissions_changed()
   dereferences its GError *, which is NULL in this case.

The fix is in step 2, to reorder emer_daemon_constructed() so that the
object is fully initialised before it calls
emer_permissions_provider_get_environment().

(You might reasonably argue that the bug is in step 3, but that is a
more invasive fix.)

https://phabricator.endlessm.com/T33305

### DIFF
--- a/daemon/emer-daemon.c
+++ b/daemon/emer-daemon.c
@@ -1480,17 +1480,17 @@ emer_daemon_constructed (GObject *object)
       g_free (network_send_path);
     }
 
-  gchar *environment =
-    emer_permissions_provider_get_environment (priv->permissions_provider);
-  schedule_upload (self, environment);
-  g_free (environment);
-
   if (priv->aggregate_tally == NULL)
     {
       priv->aggregate_tally =
         emer_aggregate_tally_new (priv->persistent_cache_directory ?: g_get_user_cache_dir ());
     }
   buffer_past_aggregate_events (self);
+
+  gchar *environment =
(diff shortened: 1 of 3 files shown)


dbnicholson (approved):
Seems reasonable.

PR #284, opened 2022-07-04
Store running timers on shutdown, add test & fix some exposed bugs

wjt:
I initially set out to fix the specific issue that running timers are not stored when the metrics daemon shuts down cleanly. In doing so, I added some tests, at which point it was relatively easy to test some other interesting cases, which turned out to fail. So there are some other changes & fixes to the timer code here.

This incorporates the test cleanups that were hanging out on #283 for a while.

https://phabricator.endlessm.com/T32730


GeorgesStavracas (approved):
Looks fantastic. Left two minor comments, but LGTM regardless of them, feel free to merge

wjt:
Thanks, I fixed those, removed another stray line, improved the commit message of the crucial commit.

PR #286, opened 2022-10-24
Add .mailmap

wjt:
I can't remember quite why I decided to do this, but I had it lying
around.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,20 @@
+Chris Chiu <chiu@endlessos.org> <chiu@endlessm.com>
+Cosimo Cecchi <cosimoc@gnome.org> <cosimo@endlessm.com>
+Dan Nicholson <dbn@endlessos.org> <nicholson@endlessm.com>
+Derek Foreman <derek@endlessm.com>
+Devin Ekins <devinj.ekins@gmail.com>
+Devin Ekins <devinj.ekins@gmail.com> <DevinJ.Ekins@gmail.com>
+Kurt von Laven <kurt.von.laven@gmail.com> <kurt@endlessm.com>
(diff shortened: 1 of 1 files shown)


.mailmap
```diff
@@ -0,0 +1,20 @@
+Chris Chiu <chiu@endlessos.org> <chiu@endlessm.com>
+Cosimo Cecchi <cosimo@endlessm.com> <cosimoc@gnome.org>
+Dan Nicholson <dbn@endlessos.org> <nicholson@endlessm.com>
+Derek Foreman <derek@endlessm.com>
+Devin Ekins <devinj.ekins@gmail.com>
+Devin Ekins <devinj.ekins@gmail.com> <DevinJ.Ekins@gmail.com>
+Kurt von Laven <kurt@endlessm.com>
```
Kurt-von-Laven, line 7: You can map me to kurt.von.laven@gmail.com since this email no longer exists. Thanks for helping maintain this project!

Kurt-von-Laven (changes requested):

.mailmap
wjt: Hi, great to hear from you! I think we missed overlapping at Endless by just a couple of months.

Kurt-von-Laven: Bad luck!

Kurt-von-Laven: As a thinly veiled excuse to meet, I am happy to share everything I think I did wrong/would have done differently in retrospect If the metrics system ever undergoes a major refactor.

dbnicholson (approved):
I guess https://git-scm.com/docs/gitmailmap? Seems nice to have the commits in this deep history point to people as they exist now.

.mailmap
wjt: That'd be great – I'll drop you an email!

PR #285, opened 2022-10-24
Miscellaneous cleanup

wjt:
While starting to think about https://phabricator.endlessm.com/T29742 I spotted a warning from Meson, and accidentally 1,600 LOC.

PR #287, opened 2022-10-25
Make test-daemon fast by mocking the clock

wjt:
Builds on #285.

Without this PR, `test-daemon` takes over 40 seconds. With this PR, it takes more like 1.5s.

This is achieved by replacing real 2-second timeouts with a mocked view of time under the control of the test suite. I have only mocked the one function necessary to make this test fast, but you could imagine going further down this road and testing things like the midnight rollover of the aggregate tally.

https://phabricator.endlessm.com/T33952


tests/daemon/test-daemon.c
```diff
@@ -1340,10 +1387,12 @@ main (gint                argc,
                    test_daemon_records_aggregates);
   ADD_DAEMON_TEST ("/daemon/retries-singular-uploads",
                    test_daemon_retries_singular_uploads);
+if (FALSE) {
```
wjt, line 1390: Uh, I guess I should fix these too before making it not a draft!

tests/daemon/mock-clock.c
```diff
@@ -0,0 +1,138 @@
...
+
+typedef struct _MockClockSource {
+  GSource parent;
+
+  MockClock *clock;
+  MockClockType type_;
+
+  /* Microseconds */
+  guint interval;
+  gint64 next_ready_time;
```
wjt, line 23: I borrowed this from eos-payg. In that project, the ready time is split into two 32-bit fields:

https://github.com/endlessm/eos-payg/blob/master/libeos-payg/fake-clock.c#L113-L114

This turns out to be because on 32-bit ARM, `GSource` has 32-bit alignment, and putting a 64-bit variable into that struct (that extends `GSource`) makes it require 64-bit alignment, and the compiler (rightly) complains.

https://github.com/endlessm/eos-payg/commit/3f0144210c6e5aa1e8f24a9a5e8df533e1be6f21

We no longer support 32-bit architectures though.

wjt: (Ignoring 32-bit makes all this code *way* shorter.)

wjt:
Jenkins runs the test suite twice so the effect of this PR is doubly visible there:

Build time drops steeply from nearly 3 minutes to a bit over 1

tests/daemon/mock-clock.c
```diff
@@ -22,13 +40,99 @@ mock_clock_class_init (MockClockClass *klass)
 {
 }
 
+static gboolean
+mock_clock_source_check (GSource *source)
+{
+  MockClockSource *self = (MockClockSource *) source;
+
+  g_debug ("%s(%p)", G_STRFUNC, source);
```
wjt, line 48: Probably should take out my stray debug statements…

tests/daemon/mock-clock.c
```diff
@@ -22,13 +40,89 @@ mock_clock_class_init (MockClockClass *klass)
...
+
+static GSource *
+mock_clock_timeout_source_new_seconds (EmerClock *self,
+                                       guint      interval)
+{
+  g_return_val_if_fail (MOCK_IS_CLOCK (self), NULL);
+
+  return mock_clock_source_new (MOCK_CLOCK (self),
+                                MOCK_CLOCK_TYPE_MONOTONIC,
+                                interval);
```
pwithnall, line 104: This is where your integer/microsecond conflation is happening. `interval` is definitely in seconds here. It’s unclear whether the `interval` argument to `mock_clock_source_new()` is in seconds or microseconds, but then `source->interval` should definitely be in microseconds.

tests/daemon/mock-clock.c
```diff
@@ -22,13 +40,89 @@ mock_clock_class_init (MockClockClass *klass)
...
 clock_iface_init (EmerClockInterface *iface,
                   gpointer            iface_data)
 {
+  iface->timeout_source_new_seconds = mock_clock_timeout_source_new_seconds;
 }
 
 static void
 mock_clock_init (MockClock *self)
 {
+  self->times[MOCK_CLOCK_TYPE_MONOTONIC] = 0x1;
```
pwithnall, line 117: Why is this initialised to 1? (And why is the 1 written in hex??)

tests/daemon/test-daemon.c
```diff
@@ -64,6 +64,9 @@
 
 #define TIMEOUT_SEC 5
 
+#define INTERVAL_SEC 3600
+#define INTERVAL_USEC (INTERVAL_SEC + G_USEC_PER_SEC)
```
pwithnall, line 68: Should this be `INTERVAL_SEC * G_USEC_PER_SEC` rather than `+`?

tests/daemon/mock-clock.c
```diff
@@ -22,13 +40,89 @@ mock_clock_class_init (MockClockClass *klass)
...
+}
+
+void
+mock_clock_advance_monotonic (MockClock *self,
+                              gint64     delta_usecs)
+{
+  g_return_if_fail (delta_usecs >= 0);
```
pwithnall, line 124: If `delta_usecs` is never allowed to be negative, would it be clearer to make it a `guint64` rather than `gint64`?

pwithnall (changes requested):
I realise this is a draft, but I thought I’d take a look while I was in the area.

I think I’ve tracked down the seconds/microseconds confusion issues. There might be something I’ve missed, but I’ve definitely found some issues which could be throwing the tests off.

tests/daemon/test-daemon.c
wjt: **Great** catch. I found the other confusion above and it's fixed in my working copy, but then had some really baffling failures in the test suite. I stared at the logs for hours before giving up!

tests/daemon/mock-clock.c
```diff
@@ -22,13 +40,89 @@ mock_clock_class_init (MockClockClass *klass)
...
+  self->next_ready_time += self->interval;
+  return callback (user_data);
+}
+
+static void
+mock_clock_source_finalize (GSource *source)
+{
+  MockClockSource *self = (MockClockSource *) source;
+
+  g_clear_object (&self->clock);
```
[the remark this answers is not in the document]
wjt: It's not a GObject `finalize`, it's a `GSourceFuncs.finalize`. I don't think these do chaining up?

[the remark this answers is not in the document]
pwithnall: aha, I skim-read it as a `GObject.finalize`, so you are correct

tests/daemon/mock-clock.c
wjt: It can't be outside the positive range of a gint64 because `times_us` is a gint64 because GLib represents such timestamps as a signed gint64. I don't think this looks any better:

```c
  g_return_if_fail (delta_us < G_MAXINT64);
  g_return_if_fail (self->times_us[MOCK_CLOCK_TYPE_MONOTONIC] <= G_MAXINT64 - (gint64) delta_us);
```

(The cast is needed because `times_us[]` is signed; otherwise the compiler gets sad.)

tests/daemon/mock-clock.c
wjt: I deleted the initialisation.

tests/daemon/mock-clock.c
wjt: I named all microsecond fields/variables/parameters as `foo_us` and fixed this up.

shared/emer-clock.c
```diff
@@ -7,3 +7,20 @@ emer_clock_default_init (EmerClockInterface *iface)
 {
 
 }
+
+guint
+emer_clock_timeout_add_seconds (EmerClock  *self,
+                                guint       interval,
```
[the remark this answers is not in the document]
wjt: I implemented this with `g_source_set_name` for now.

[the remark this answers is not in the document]
pwithnall: Hopefully the following?
```c
#if GLIB_CHECK_VERSION(2, 70, 0)
  g_source_set_static_name (source, static_name);
#else
  g_source_set_name (source, static_name);
#endif

tests/daemon/mock-clock.c
pwithnall: Could `times_us` also be `guint64`?

wjt: It could, but there is also code in `emer-daemon` which retrieves the current monotonic time, which is a `gint64`. Currently it gets the real time, but I would like later to mock that, too. I think it makes sense to be consistent with the way the regular GLib functions represent time.

pwithnall: > I think it makes sense to be consistent with the way the regular GLib functions represent time.

Sure, that’s a valid argument and I’m not going to push further. Though I will point out that, as far as I am aware, it’s a historical accident that the GLib functions use `gint64` rather than `guint64`.

Issue #292, opened 2023-03-21
g_variant_new_string: assertion 'string != NULL' failed


mcatanzaro:
Using commit ab67c9347cdc52be6a8f63d8bb4e9c43584d3a85 on Fedora 38, when trying to upload metrics I see `g_variant_new_string: assertion 'string != NULL' failed`. I got this backtrace:

```
#0  g_logv
    (log_domain=0x7f46a9ca20be "GLib", log_level=G_LOG_LEVEL_CRITICAL, format=<optimized out>, args=args@entry=0x7ffcf17b3120) at ../glib/gmessages.c:1422
#1  0x00007f46a9c45fb3 in g_log (log_domain=<optimized out>, log_level=<optimized out>, format=<optimized out>)
    at ../glib/gmessages.c:1460
#2  0x00007f46a9c85f82 in g_variant_new_string (string=string@entry=0x0) at ../glib/gvariant.c:1269
#3  0x00007f46a9c87578 in g_variant_valist_new_nnp (str=str@entry=0x7ffcf17b3328, ptr=0x0) at ../glib/gvariant.c:4870
#4  0x00007f46a9c87b5e in g_variant_valist_new_leaf (app=0x7ffcf17b3350, str=0x7ffcf17b3328)
    at ../glib/gvariant.c:5065
#5  0x00007f46a9c879f3 in g_variant_valist_new (str=str@entry=0x7ffcf17b3328, app=app@entry=0x7ffcf17b3350)
    at ../glib/gvariant.c:5299
#6  0x00007f46a9c88009 in g_variant_new_va (format_string=<optimized out>, 
    format_string@entry=0x5623b78a3730 "(xxs@a{ss}ya(aysxmv)a(ayssumv))", endptr=endptr@entry=0x0, app=app@entry=0x7ffcf17b3350) at ../glib/gvariant.c:5475
#7  0x00007f46a9c8814e in g_variant_new
    (format_string=format_string@entry=0x5623b78a3730 "(xxs@a{ss}ya(aysxmv)a(ayssumv))") at ../glib/gvariant.c:5410
#8  0x00005623b789f40c in create_request_body
    (error=0x7ffcf17b34d8, num_buffer_events=<synthetic pointer>, num_stored_events=<synthetic pointer>, token=0x7ffcf17b34d0, max_bytes=<optimized out>, self=0x5623b8a66cb0) at ../daemon/emer-daemon.c:779
[...]

PR #295, opened 2023-03-22
daemon: Avoid criticals on non-EOS where image version isn't set

kalev:
Add fallbacks to read OSTREE_VERSION for Silverblue, and VERSION_ID for other distros that aren't image based.

Fixes https://github.com/endlessm/eos-event-recorder-daemon/issues/292

### DIFF
--- a/daemon/emer-image-id-provider.c
+++ b/daemon/emer-image-id-provider.c
@@ -94,6 +94,9 @@ emer_image_id_provider_get_version (void)
   if (image_version == NULL)
     image_version = get_image_version_for_path (EOS_IMAGE_VERSION_ALT_PATH);
 
+  if (image_version == NULL)
+    image_version = g_strdup ("unknown");
+
   return image_version;
 }
 


kalev:
Note that this is untested.

mcatanzaro:
We've both tested it now, and it works.

Do you think it's semantically fair to consider ostree version to be the "emer_image_id"? I guess it's fine?

wjt:
> Do you think it's semantically fair to consider ostree version to be the "emer_image_id"? I guess it's fine?

What are some examples of these fields?

This string is meant to represent the distribution channel. It is in a particular, Endless-specific format. I think you may find that Azafea will fail to import the event because it will try to parse it according to that format:

https://github.com/endlessm/azafea/blob/fcdb9ab1907f82954986931efcfc3adf8de5656a/azafea/event_processors/endless/image.py#L57-L104

Semantically the fields are:

- image_product: the distribution channel (for us this is something like `eos` for the download version, `eosoem` for OEM installs, `eoscustom` for installation images created by third parties, etc.)
- image_branch: the OS branch. In Fedora terms this would be something like f28, fc30
- image_arch: e.g. arm64, amd64
- image_platform: for the amd64 arch this is (almost) always amd64, but for ARM platforms where you need different install media for different hardware this is something like `rpi4`, `pinebookpro`
- image_timestamp: the build timestamp of the image
- image_personality: for us this something like `base` (minimum install), `en`, `pt_BR`, or the name of a particular deployment – for instance if you look at https://www.endlessos.org/partners, the personality for the deployment in Oaxaca is `oaxaca`, Teach For America is `tfa`, etc.

This is meant to relate to the original installation, not the current OS version. Each event is separately stamped with the current OS version.

These fields get parsed on the server and are the primary way that you filter data (along with time) so that you can answer questions like “what apps do users that bought OEM laptops preinstalled with Endless OS use?”.

So I think you're going to want to conjure something up in the same string format (for the time being). I would claim that you will want the different spins/flavours on https://getfedora.org/ to be tagged differently, so e.g. one of the fields will want to distinguish `silverblue` from `workstation`, and it would be useful to put the originally-installed Fedora branch (if such information exists!) into the branch field.

mcatanzaro:
> We've both tested it now, and it works.

Well, I should clarify: it fixes #292, but indeed azafea rejects the events.

kalev:
Thanks! I updated it to fall back to "unknown" as per Matrix discussion. We'll still need to figure out how to construct this correctly but at least unknown should make Azafea not reject the submission.

wjt:
Hey, Jenkins! add to whitelist!
Hey, humans! We need to change our Jenkins configuration to use less problematic terminology for its PR commands.

PR #298, opened 2024-04-19
Add a tool to print buffered aggregates to the CLI

wjt:
I have written a half-baked version of this useful tool many times. Let's put it in the source tree.

While writing some tests, I found some minor misuses of the sqlite3 API which lead to memory leaks in the error paths of basically every function. So, draft for that reason.

https://phabricator.endlessm.com/T35331


daemon/emer-aggregate-tally.c
```diff
@@ -178,16 +187,19 @@ delete_tally_entries (EmerAggregateTally  *self,
...
-    }
+  gboolean ret =
+      CHECK (sqlite3_prepare_v2 (self->db, query->str, -1, &stmt, NULL)) &&
+      CHECK (sqlite3_step (stmt));
 
-  return TRUE;
+  if (ret)
+    ret = CHECK (sqlite3_finalize (stmt));
+  else
+    sqlite3_finalize (stmt);
```
wjt, line 197: This repeated pattern feels really ugly but I can't think of a way to make it more concise without adding yet more macros. The `CHECK` macro is already a little bit too magic for my tastes today – and I wrote it!

daemon/emer-aggregate-tally.c
```diff
@@ -178,16 +187,19 @@ delete_tally_entries (EmerAggregateTally  *self,
     }
   g_string_append (query, ");");
 
-  if (!CHECK (sqlite3_prepare_v2 (self->db, query->str, -1, &stmt, NULL)) ||
-      !CHECK (sqlite3_step (stmt)) ||
-      !CHECK (sqlite3_reset (stmt)) ||
```
wjt, line 183: `sqlite3_reset()` followed by `sqlite3_finalize()` is redundant, which is why I have removed it.

daemon/emer-aggregate-tally.c
wjt: I tried writing another macro but it actually made it worse.

tools/meson.build
```diff
@@ -17,6 +17,21 @@ executable('print-persistent-cache',
...
+]
+
+executable('print-aggregates',
+    print_aggregates_sources,
+    dependencies: [
+        emer_required_modules,
+        emer_shared_dep,
+    ],
+    include_directories: include_directories('../daemon'),
+    install: false,
```
wjt, line 32: It would probably be more useful to install this (perhaps under a different name) but I'd want to somehow make it clear that it's a debugging tool rather than something that can be relied upon in scripts, to reserve the right to change it at any time.

dylanmccall:
I read through this and I didn't spot anything amiss. Those sqlite3 docs are … unique. I'll play with this locally, next, just to be sure.

dylanmccall (approved):

